PR #38, opened 2020-07-28
fix(errors): improve error messages

JCustin:
Rewrote error messages with a consistent structure to explain said errors, as well as a transition to the administrator's contact information. The administrator's information will needed to be plugged in by a more knowledgeable contributor.

Note that because I am not familiar with Go, some of the error messages may have an incorrect explanation for the error. I encourage a double check on the terminology and logic explained in my messages. If there are any errors, please correct the terminology and logic while retaining the sentence structure.

PR #40, opened 2020-07-29
fix: remove bracket

fgerschwiler:


fforootd:
this resolves a problem with #38

PR #51, opened 2020-09-07
fix: handle single `aud` string claim, extract en/decoder interface, comments

livio-a:


pkg/op/authrequest_test.go
```diff
@@ -3,66 +3,139 @@ package op_test
...
-	// 		// require.Equal(t, clientID, client.)
-	// 	}
-	// }
-	// testErr := func(t *testing.T, expected error) errorHandler {
-	// 	return func(w http.ResponseWriter, r *http.Request, authReq *oidc.AuthRequest, err error) {
-	// 		require.Equal(t, expected, err)
-	// 	}
-	// }
+//
+//func TestAuthorize(t *testing.T) {
```
adlerhurst, line 21:
```suggestion
//TOOD: tests will be implemented in branch for service accounts
//func TestAuthorize(t *testing.T) {
```

pkg/op/error.go
```diff
@@ -24,11 +24,17 @@ var (
 	}
 	ErrInvalidRequestRedirectURI = func(description string) *OAuthError {
 		return &OAuthError{
-			ErrorType:        InvalidRequest,
+			ErrorType:        InvalidRequestUri,
```
adlerhurst, line 27:
```suggestion
			ErrorType:        InvalidRequestURI,
```

adlerhurst (approved):

PR #69, opened 2020-10-28
fix: allow additional scopes

fgerschwiler:


pkg/op/token.go
```diff
@@ -107,9 +108,8 @@ func CreateIDToken(ctx context.Context, issuer string, authReq AuthRequest, vali
 		claims.SetAccessTokenHash(atHash)
 		scopes = removeUserinfoScopes(scopes)
 	}
-	if !additonalScopes {
-		scopes = removeAdditionalScopes(scopes)
-	}
+	scopes = restictAdditionalScopesFunc(scopes)
```
livio-a, line 111: just noticed: we should call `restictAdditionalScopesFunc` before checking if an accessToken is issued and `removeUserinfoScopes` to ensure that none of these scope are asserted into claims:

```
...
	scopes := restictAdditionalScopesFunc(authReq.GetScopes())
	if accessToken != "" {
		atHash, err := oidc.ClaimHash(accessToken, signer.SignatureAlgorithm())
		if err != nil {
			return "", err
		}
		claims.SetAccessTokenHash(atHash)
		scopes = removeUserinfoScopes(scopes)
	}
	if len(scopes) > 0 {
...
```

livio-a: and i believe removeAdditionalScopes below is not used anymore

livio-a (approved):

PR #152, opened 2022-02-07
feat: Allow the use of custom discovery endpoint

ydrisR:
Thanks for this solid library.
On some occasions public services will expose OAuth2/OIDC endpoint on custom uri which are different from the standard patern you use in the library. I'd like to propose a new constructor to create a RelyingParty instances based on custom endpoints, different than what would be computed by the discovery method.

### DIFF
--- a/pkg/client/client.go
+++ b/pkg/client/client.go
@@ -26,8 +26,13 @@ var (
 )
 
 //Discover calls the discovery endpoint of the provided issuer and returns its configuration
-func Discover(issuer string, httpClient *http.Client) (*oidc.DiscoveryConfiguration, error) {
+//It accepts an optional argument "wellknownUrl" which can be used to overide the dicovery endpoint url
+func Discover(issuer string, httpClient *http.Client, wellKnownUrl ...string) (*oidc.DiscoveryConfiguration, error) {
+
 	wellKnown := strings.TrimSuffix(issuer, "/") + oidc.DiscoveryEndpoint
+	if len(wellKnownUrl) == 1 && wellKnownUrl[0] != "" {
+		wellKnown = wellKnownUrl[0]
+	}
 	req, err := http.NewRequest("GET", wellKnown, nil)
 	if err != nil {
 		return nil, err
--- a/pkg/client/rp/relaying_party.go
+++ b/pkg/client/rp/relaying_party.go
@@ -69,11 +69,12 @@ var (
 )
 
 type relyingParty struct {
-	issuer      string
-	endpoints   Endpoints
-	oauthConfig *oauth2.Config
-	oauth2Only  bool
-	pkce        bool
+	issuer            string
+	DiscoveryEndpoint string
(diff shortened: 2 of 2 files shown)


fforootd:
@ydrisR thanks for the PR we will have a look on this in the next few day, hopefully your are not in a hurry 😁

fforootd:
@ydrisR one question I have is... you want to configure explicit endpoints instead of using the once provided by the oidc discovery endpoint. Or do you want to configure a oidc rp without an oidc discovery endpoint on the op side?

ydrisR:
> @ydrisR one question that I have is. Do you want to configure explicit endpoints instead of using the once provided by the oidc discovery endpoint, or do you want to configure a oidc rp without an oidc discovery endpoint on the op side?

The problem I faced is that my oidc provider decided to use a discovery endpoint which does not follow the standard patern you implemented in the library.
Here is the discovery endoint url: https://auth.bas.esw.esante.gouv.fr/auth/realms/esante-wallet/.well-known/wallet-openid-configuration 
Notice the extra "wallet-" which prevent method `oidc.client.Discover()` to work properly :(

I decided to create a new method to instantiate a new `RelyingParty` from a custom `DiscoveryConfiguration` as I thought it would give more flexibility: you can build the DiscoveryConfiguration manually or from a discovery endpoint, it's up to the developer.

For my specific situation I think we could address the problem with a simpler solution: if there was a way to set/override the constant `oidc.DiscoveryEndpoint` that would make the job.

fforootd:
> The problem I faced is that my oidc provider decided to use a discovery endpoint which does not follow the standard patern you implemented in the library.

Ah I see 😁 
What an odd design choice. I mean the oidc discovery would allow [prefixed paths for multi tenancy reasons](https://openid.net/specs/openid-connect-discovery-1_0.html#ProviderConfigurationRequest).

@livio-a where do you stand on this. IMO we should allow to overwrite the oidc discovery url.

oliverpool:
Another possible approach for this problem, would be to add a `*oidc.DiscoveryConfiguration` field on the `rp` struct, which could be set by an `Option` `WithDiscoveryConfiguration`.

In `NewRelyingPartyOIDC`, if this pointer is not nil, then it is used instead of calling `client.Discover`.

ydrisR:
I like both solutions. Don't hesitate to ping me if you want me to implement one or the other :)

livio-a:
Hey @ydrisR 

Thanks for great feedback and the issue. I have thought about it and agree with @fforootd:
IMO the most useful for your scenario would be to create an option to overwrite the discovery path. 

I think that providing a static configuration might make sense in very rare cases (not yours). If the OP only changes a single configuration, you'll have to change it as well and even recompile. Which kind of breaks the idea of the discovery endpoint.

Would you mind changing the PR to just overwrite the discovery path?

livio-a (approved):

livio-a:
Thanks for your contribution @ydrisR !

PR #168, opened 2022-04-06
feat(rp): provide key by data (not only path) for jwt profile

livio-a:


morigs (approved):
Awesome!

pkg/client/rp/relaying_party.go
```diff
@@ -233,6 +233,9 @@ func WithVerifierOpts(opts ...VerifierOption) Option {
 	}
 }
 
+// WithClientKey specifies the path to the key.json to be used for the JWT Profile Client Authentication on the token endpoint
+//
+//deprecated: use WithJWTProfile(SignerFromKeyPath(path)) instead
```
adlerhurst, line 238: I would replace the function body with the new option.

livio-a: fixed

PR #172, opened 2022-04-15
feat(op): implemented support for client_credentials grant

Place1:
This PR adds support for the client_credentials grant type following the RFC here: https://datatracker.ietf.org/doc/html/rfc6749#section-4.4

### DIFF
--- a/pkg/op/op.go
+++ b/pkg/op/op.go
@@ -230,7 +230,7 @@ func (o *openidProvider) GrantTypeJWTAuthorizationSupported() bool {
 }
 
 func (o *openidProvider) GrantTypeClientCredentialsSupported() bool {
-	_, ok := o.storage.(ClientCredentialsXXX)
+	_, ok := o.storage.(ClientCredentialsStorage)
 	return ok
 }
 
--- a/pkg/op/op.go
+++ b/pkg/op/op.go
@@ -229,6 +229,11 @@ func (o *openidProvider) GrantTypeJWTAuthorizationSupported() bool {
 	return true
 }
 
+func (o *openidProvider) GrantTypeClientCredentialsSupported() bool {
+	_, ok := o.storage.(ClientCredentialsXXX)
+	return ok
+}
+
 func (o *openidProvider) IntrospectionAuthMethodPrivateKeyJWTSupported() bool {
 	return true
 }
--- a/pkg/op/storage.go
+++ b/pkg/op/storage.go
@@ -27,6 +27,10 @@ type AuthStorage interface {
 	GetKeySet(context.Context) (*jose.JSONWebKeySet, error)
 }
 
+type ClientCredentialsXXX interface {
+	ClientCredentialsTokenRequest(ctx context.Context, clientID string, scopes []string) (TokenRequest, error)
+}
+
 type OPStorage interface {
 	GetClientByClientID(ctx context.Context, clientID string) (Client, error)
 	AuthorizeClientIDSecret(ctx context.Context, clientID, clientSecret string) error
--- a/pkg/op/storage.go
+++ b/pkg/op/storage.go
@@ -27,7 +27,7 @@ type AuthStorage interface {
 	GetKeySet(context.Context) (*jose.JSONWebKeySet, error)
 }
 
-type ClientCredentialsXXX interface {
+type ClientCredentialsStorage interface {
 	ClientCredentialsTokenRequest(ctx context.Context, clientID string, scopes []string) (TokenRequest, error)
 }
 
--- a/pkg/op/token_client_credentials.go
+++ b/pkg/op/token_client_credentials.go
@@ -1,0 +1,120 @@
+package op
+
+import (
+	"context"
+	"net/http"
+	"net/url"
+
+	httphelper "github.com/caos/oidc/pkg/http"
+	"github.com/caos/oidc/pkg/oidc"
+)
+
+//ClientCredentialsExchange handles the OAuth 2.0 client_credentials grant, including
+//parsing, validating, authorizing the client and finally returning a token
+func ClientCredentialsExchange(w http.ResponseWriter, r *http.Request, exchanger Exchanger) {
+	request, err := ParseClientCredentialsRequest(r, exchanger.Decoder())
+	if err != nil {
+		RequestError(w, r, err)
+	}
+
+	validatedRequest, client, err := ValidateClientCredentialsRequest(r.Context(), request, exchanger)
+	if err != nil {
+		RequestError(w, r, err)
+		return
+	}
+
+	resp, err := CreateClientCredentialsTokenResponse(r.Context(), validatedRequest, exchanger, client)
+	if err != nil {
+		RequestError(w, r, err)
+		return
+	}
+
+	httphelper.MarshalJSON(w, resp)
+}
+
+//ParseClientCredentialsRequest parsed the http request into a oidc.ClientCredentialsRequest
+func ParseClientCredentialsRequest(r *http.Request, decoder httphelper.Decoder) (*oidc.ClientCredentialsRequest, error) {
+	err := r.ParseForm()
+	if err != nil {
+		return nil, oidc.ErrInvalidRequest().WithDescription("error parsing form").WithParent(err)
+	}
+
+	request := new(oidc.ClientCredentialsRequest)
+	err = decoder.Decode(request, r.Form)
+	if err != nil {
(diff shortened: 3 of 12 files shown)


pkg/op/op.go
```diff
@@ -229,6 +229,11 @@ func (o *openidProvider) GrantTypeJWTAuthorizationSupported() bool {
 	return true
 }
 
+func (o *openidProvider) GrantTypeClientCredentialsSupported() bool {
+	_, ok := o.storage.(ClientCredentialsXXX)
```
Place1, line 233: if the storage provider implements _ClientCredentialsXXX_ then the grant type is supported.

pkg/op/storage.go
```diff
@@ -27,6 +27,10 @@ type AuthStorage interface {
 	GetKeySet(context.Context) (*jose.JSONWebKeySet, error)
 }
 
+type ClientCredentialsXXX interface {
```
Place1, line 30: I've added this new method that the storage provider can implement.

The _ClientCredentialsXXX_ interface is separate from the OPStorage and AuthStorage interfaces so that this PR doesn't introduce a breaking change.

The method is used to allow the storage provider (library consumer) to control the subject/aud/scopes on the access token created during the flow.

_more comments on this below_

pkg/op/token_client_credentials.go
```diff
@@ -0,0 +1,115 @@
...
+	if err != nil {
+		RequestError(w, r, err)
+		return
+	}
+
+	httphelper.MarshalJSON(w, resp)
+}
+
+//ParseClientCredentialsRequest parsed the http request into a oidc.ClientCredentialsRequest
+func ParseClientCredentialsRequest(r *http.Request, decoder httphelper.Decoder) (*oidc.ClientCredentialsRequest, error) {
```
Place1, line 36: This function parses the http request into a ClientCredentialsRequest struct.

It supports the client id/secret as a basic auth header or post body.

Basic auth is what the RFC documents to use. I couldn't actually find any text in the RFC that says post parameters should be supported as well; but it seems very common from reading online.

pkg/op/token_client_credentials.go
```diff
@@ -0,0 +1,115 @@
...
+
+	tokenRequest, err := storage.ClientCredentialsTokenRequest(ctx, request.ClientID, request.Scope)
+	if err != nil {
+		return nil, nil, err
+	}
+
+	return tokenRequest, client, nil
+}
+
+func AuthorizeClientCredentialsClient(ctx context.Context, request *oidc.ClientCredentialsRequest, exchanger Exchanger) (Client, error) {
```
Place1, line 87: The client is authorized by calling the existing "AuthorizeClientIDSecret" OPStorage interface method.

pkg/op/token_client_credentials.go
```diff
@@ -0,0 +1,115 @@
...
+
+	if !ValidateGrantType(client, oidc.GrantTypeClientCredentials) {
+		return nil, oidc.ErrUnauthorizedClient()
+	}
+
+	return client, nil
+}
+
+func CreateClientCredentialsTokenResponse(ctx context.Context, tokenRequest TokenRequest, creator TokenCreator, client Client) (*oidc.AccessTokenResponse, error) {
+	accessToken, _, validity, err := CreateAccessToken(ctx, tokenRequest, AccessTokenTypeJWT, creator, client, "")
```
Place1, line 105: Originally I wanted to call [CreateTokenResponse](https://github.com/Place1/oidc/blob/support-for-client-credentials-grant/pkg/op/token.go#L25-L26) which is used by the current authorization_code flow implementation; but the function assumes there should be an IDToken in the response.

Given a client_credentials request doesn't have an IDToken I've had to make use of "CreateAccessToken" instead.

Place1:
The current "ClientCredentialsXXX" interface is named like this because i'm not sure what the interface with the library consumer should be exactly.

Is this something you'd like to help decide?

I'd like to get the interface right before I spend time writing tests.

fforootd:
Thanks a lot for this PR and the work done!

We should have time for looking into this next, week.

pkg/op/token_client_credentials.go
```diff
@@ -0,0 +1,115 @@
...
+
+	return request, nil
+}
+
+//ValidateClientCredentialsRequest validates the refresh_token request parameters including authorization check of the client
+//and returns the data representing the original auth request corresponding to the refresh_token
+func ValidateClientCredentialsRequest(ctx context.Context, request *oidc.ClientCredentialsRequest, exchanger Exchanger) (TokenRequest, Client, error) {
+	storage, ok := exchanger.Storage().(ClientCredentialsXXX)
+	if !ok {
+		return nil, nil, oidc.ErrRequestNotSupported().WithDescription("client_credentials grant not supported")
```
livio-a, line 71:
```suggestion
		return nil, nil, oidc.ErrUnsupportedGrantType().WithDescription("client_credentials grant not supported")
```

Place1:
> ClientCredentialStorage

I'll put this in the PR unless there's another suggestion :D

Place1:
> One question though: Wouldn't the addition of GrantTypeClientCredentialsSupported() bool to the Configuration and Exchanger interface be such a breaking change?

I wasn't sure if the Exchanger and Configuration interfaces were public for library consumers to use.

I was under the impression that only the `op.Storage` interface was expected to be used by consuming applications via the [NewOpenIDProvider](https://github.com/Place1/oidc/blob/support-for-client-credentials-grant/pkg/op/op.go#L120-L121) function.

I've added the `GrantTypeClientCredentialsSupported` method to these interfaces to be consistent with the other `xxxSupported` methods.

I'm not sure what's best to do here. I'm probably lacking context about the broader use-cases for the library :)

livio-a:
> I wasn't sure if the Exchanger and Configuration interfaces were public for library consumers to use.
> 
> I was under the impression that only the `op.Storage` interface was expected to be used by consuming applications via the [NewOpenIDProvider](https://github.com/Place1/oidc/blob/support-for-client-credentials-grant/pkg/op/op.go#L120-L121) function.
> 
> I've added the `GrantTypeClientCredentialsSupported` method to these interfaces to be consistent with the other `xxxSupported` methods.
> 
> I'm not sure what's best to do here. I'm probably lacking context about the broader use-cases for the library :)
> 
> Also, the type cast/check you mentioned ([this one](https://github.com/Place1/oidc/blob/support-for-client-credentials-grant/pkg/op/token_request.go#L39-L40)) seems weird. The method is already on the Exchanger interface so a cast is not required; unless the cast is part of the business logic itself in which case the `GrantTypeJWTAuthorizationSupported` should probably be defined on the `JWTAuthorizationGrantExchanger` interface instead.

You're basically right. The main purpose is to implement the `op.Storage` interface and let the library handle the oidc / protocol part as far as possible.
However, you could also reuse the functions / interfaces to create an own implementation of the OpenID Provider (interface).
But I don't think anyone is currently doing this.
About the cast/check: You're absolutely right! With the current Exchanger interface it's completely unnecessary. Going to fix that later on.

So I'm ok with the extension of the two interfaces / current PR. @fforootd any objections?

fforootd:
I think we should update the table in the readme with the supported flows. 

Otherwise it looks good to me.

Place1:
Hello, just wondering what's needed from me to get this moving. I'd really like to see this feature merged :D

livio-a:
Sorry, simply forgot this.

We moved the repo to the zitadel organisation in the meantime.
Could you rebase your PR / update the paths in your PR to `github.com/zitadel/oidc` please?

Afterwards, the checks should be ok and will be able to merge.

Place1:
Thanks for taking a look @livio-a 

I've rebased and updated import paths. Did you need me to squash as well?

livio-a:
> Thanks for taking a look @livio-a
> 
> I've rebased and updated import paths. Did you need me to squash as well?

We'll squash into main, no need on your side.
But apparently some tests failed because of the updated `Configuration` interface.
These tests use a generated mock. I'll have to generate them automatically in the future. In the meantime you'll either have to quickly run `go generate ./pkg/op/mock/generate.go` to have the latests mocks for the interfaces. 
A quick `go test ./pgk/...` should then be ok.
Sorry again.

livio-a:
@Place1 thanks again and sorry for the delay!

Place1:
Thanks so much :D

fforootd:
@Place1 thank you for the PR.

I am intrigued to send you over some ZITADEL merch (as soon as we got the new stuff). Contact me at florian@caos.ch if you are interested 😁

PR #179, opened 2022-05-11
feat(rp): Adding end_session endpoint to relaying party interface

zuchi:
In my case here I had to have a way to know what would be the endpoint for signing out. 
I realized that the library doesn't expose this endpoint in the rp package. So what I did was add a new function in the interface and create a new field to expose the end session endpoint. 

I'm opening here this Pull request just for you all to do an analysis and see if it breaks any concepts regarding the OpenID Connect.

If it breaks anything (in the concept) please let me know why.

By the way, thank you all for maintaining this project, that is really nice.

### DIFF
--- a/pkg/client/rp/relaying_party.go
+++ b/pkg/client/rp/relaying_party.go
@@ -50,6 +50,9 @@ type RelyingParty interface {
 	//Signer is used if the relaying party uses the JWT Profile
 	Signer() jose.Signer
 
+	//GetEndSessionEndpoint returns the endpoint to sign out on a IDP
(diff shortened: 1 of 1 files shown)


livio-a:
Hey @zuchi 

Thanks for your feedback and this PR of course 😃 
It certainly makes sense to make the endpoint available to the RP. We must have missed that, because we didn't need it in our use case.
I can't see anything breaking and I'd like to merge it.

livio-a (approved):

livio-a:
@zuchi thanks again for the contribution!

PR #185, opened 2022-06-17
fix: state and auth code response encoding

### LINKED ISSUE #183: missing state in redirect of implicit flow
**Describe the bug**

as described here: https://discord.com/channels/927474939156643850/927866013545025566/981120399658082334

> if I create the default quickstart react app, and using either Zitadel v1, or v2, setup a client to allow id_token, then configure the app to use id_token response type instead of code, the app doesn't work because the redirect from Zitadel doesn't include the state that the client 

**Expected behaviour**

The redirect should include the state field.

livio-a:
close #183

### DIFF
--- a/pkg/op/auth_request.go
+++ b/pkg/op/auth_request.go
@@ -465,18 +465,41 @@ func BuildAuthRequestCode(authReq AuthRequest, crypto Crypto) (string, error) {
 //AuthResponseURL encodes the authorization response (successful and error) and sets it as query or fragment values
 //depending on the response_mode and response_type
 func AuthResponseURL(redirectURI string, responseType oidc.ResponseType, responseMode oidc.ResponseMode, response interface{}, encoder httphelper.Encoder) (string, error) {
-	params, err := httphelper.URLEncodeResponse(response, encoder)
+	uri, err := url.Parse(redirectURI)
 	if err != nil {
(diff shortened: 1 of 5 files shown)


pkg/op/auth_request.go
```diff
@@ -465,18 +465,41 @@ func BuildAuthRequestCode(authReq AuthRequest, crypto Crypto) (string, error) {
...
+func setFragment(uri *url.URL, params url.Values) string {
+	uri.Fragment = params.Encode()
+	return uri.String()
+}
+
+func mergeQueryParams(uri *url.URL, params url.Values) string {
+	queries := uri.Query()
+	for param, values := range params {
+		for _, value := range values {
+			queries.Set(param, value)
```
adlerhurst, line 500: do we have to document that we overwrite predefined the queries with our params?

livio-a: i'll change that to `.Add()`
currently we don't have a param with multiple values, but if we ever do the current implementation would only set the last value 🤦 

and either way (overwrite or add) it's the clients issue to solve it...
if we overwrite it, the client will not get the desired state/code/...
if we append,  it's again the client issue to be able to parse multiple values for the same param

hifabienne (approved):

PR #188, opened 2022-06-27
chore: fix typo in filename

muir:
I keep wanting to type relaying_party too.


fforootd:
Hi @muir

Thanks for this PR! You are totally right 😁 that's a typo.

@livio-a can you have a look at this.

livio-a:
> I keep wanting to type relaying_party too.

took us a while to notice that we even created the interface like that 🤦 
so thanks for the PR 😃

livio-a (approved):

PR #191, opened 2022-06-29
chore: add some docs to NewOpenIDProvider()

muir:
I had some difficulty understanding how to use the op library.  I've added a comment that would have helped me.

Is my comment accurate?


livio-a (approved):
@muir your comment is absolutely correct and thank you for contributing again

PR #193, opened 2022-07-07
chore: add comments documenting Storage and AuthStorage

muir:
Working to implement the Storage interfaces, I found myself puzzled over what I could count on for `TokenRequest` beyond it's minimal interface.  It's minimal interface isn't rich enough for me to implement `CreateAccessToken`.   I've tried to address that with a comment, but a change in the API would be better (yeah,I know that would break backwards compatibility; sometimes that needs to happen).

Are my comments accurate?    What flow results in creating an access token after decoding a JWT?

Thanks

Note: I'm building a federated identity server so requests to my OP will trigger an outbound request as an OIDC RP or as a SAML SP depending on the SSO setup for that particular user.  After I get things working, I expect to try to make a change so that the OP /callback can be combined into the success handlers for the client authentication callbacks and thus avoid an intra-server redirect.

### DIFF
--- a/pkg/op/storage.go
+++ b/pkg/op/storage.go
@@ -20,9 +20,6 @@ type AuthStorage interface {
 	//
 	// * TokenRequest as returned by ClientCredentialsStorage.ClientCredentialsTokenRequest,
 	//
-	// * RefreshTokenRequest as returned by AuthStorage.TokenRequestByRefreshToken
-	//   (CreateAccessAndRefreshTokens will also be called)
-	//
 	// * AuthRequest as returned by AuthRequestByID or AuthRequestByCode (above)
 	//
 	// * *oidc.JWTTokenRequest from a JWT that is the assertion value of a JWT Profile


pkg/op/storage.go
```diff
@@ -16,12 +16,22 @@ type AuthStorage interface {
 	SaveAuthCode(context.Context, string, string) error
 	DeleteAuthRequest(context.Context, string) error
 
-	CreateAccessToken(context.Context, TokenRequest) (string, time.Time, error)
-	CreateAccessAndRefreshTokens(ctx context.Context, request TokenRequest, currentRefreshToken string) (accessTokenID string, newRefreshToken string, expiration time.Time, err error)
-	TokenRequestByRefreshToken(ctx context.Context, refreshToken string) (RefreshTokenRequest, error)
+	// The TokenRequest parameter of CreateAccessToken can be any of:
+	// - TokenRequest as returned by ClientCredentialsStorage.ClientCredentialsTokenRequest
+	// - RefreshTokenRequest as returned by AuthStorage.TokenRequestByRefreshToken
+	// - AuthRequest as returned one of the AuthStorage methods above
```
livio-a, line 22: the AuthRequest can also be returned by the AuthStorage.AuthRequestByID and AuthStorage.AuthRequestByCode (used for the authorization code flow)

pkg/op/storage.go
```diff
@@ -16,12 +16,22 @@ type AuthStorage interface {
...
+	// - *oidc.JWTTokenRequest created by decoding a JWT
```
livio-a, line 23: the JWT is the `assertion` value of a JWT Profile Grant: https://datatracker.ietf.org/doc/html/rfc7523#section-2.1

pkg/op/storage.go
```diff
@@ -16,12 +16,22 @@ type AuthStorage interface {
...
+	CreateAccessToken(context.Context, TokenRequest) (accessTokenID string, expiration time.Time, err error)
+
+	// The TokenRequest parameter of CreateAccessAndRefreshTokens can be any of:
+	// - TokenRequest as returned by ClientCredentialsStorage.ClientCredentialsTokenRequest
+	// - RefreshTokenRequest as returned by AuthStorage.TokenRequestByRefreshToken
+	// - AuthRequest as returned one of the AuthStorage methods above
```
livio-a, line 29: the AuthRequest can also be returned by the AuthStorage.AuthRequestByID and AuthStorage.AuthRequestByCode (used for the authorization code flow) which requested `offline_access` scope and registered the `refresh_token` grant type in advance

livio-a (commented):
sorry for the late response and thank you very much for the continuous work on documentation and small fixes
i've made some comments and additions to yours

BTW I'm open to changes to the interface. We've already created a v2 pre-release because we needed a way to dynamically get the issuer based on the host in ZITADEL (https://github.com/zitadel/oidc/tree/dynamic-issuer)

pkg/op/storage.go
```diff
@@ -16,12 +16,22 @@ type AuthStorage interface {
 	SaveAuthCode(context.Context, string, string) error
 	DeleteAuthRequest(context.Context, string) error
 
-	CreateAccessToken(context.Context, TokenRequest) (string, time.Time, error)
-	CreateAccessAndRefreshTokens(ctx context.Context, request TokenRequest, currentRefreshToken string) (accessTokenID string, newRefreshToken string, expiration time.Time, err error)
-	TokenRequestByRefreshToken(ctx context.Context, refreshToken string) (RefreshTokenRequest, error)
+	// The TokenRequest parameter of CreateAccessToken can be any of:
+	// - TokenRequest as returned by ClientCredentialsStorage.ClientCredentialsTokenRequest
+	// - RefreshTokenRequest as returned by AuthStorage.TokenRequestByRefreshToken
```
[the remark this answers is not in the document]
muir: So CreateAccessAndRefreshTokens and CreateAccessToken will _both_ be called?   That seems redundant with respect to the access tokens.

pkg/op/storage.go
muir: noted

pkg/op/storage.go
muir: reworded.  I thought I had that covered.

pkg/op/storage.go
muir: Thanks, incorporated.

muir:
> sorry for the late response and thank you very much for the continuous work on documentation and small fixes i've made some comments and additions to yours
> 
> BTW I'm open to changes to the interface. We've already created a v2 pre-release because we needed a way to dynamically get the issuer based on the host in ZITADEL (https://github.com/zitadel/oidc/tree/dynamic-issuer)

Great!

I'm not yet confident enough to make the changes to the interface.  My suggestion here would be to split up these two methods into ones that take more specific parameters and drop the `op.TokenRequest` type.

pkg/op/storage.go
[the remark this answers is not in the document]
livio-a: sorry I expressed myself quite unfortunate... I meant that the RefreshTokenRequest always calls CreateAccessAndRefreshTokens and not CreateAccessToken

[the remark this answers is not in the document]
muir: Got it.  Changed. 

I think everything else is also addressed.

livio-a (approved):

PR #197, opened 2022-07-21
chore: add enumer for iota-defined types

muir:
This adds an enumer for the types that are defined with `iota`.   The intent is to make it easier for implementors to load and store the related data types.

```
//go:generate go get github.com/dmarkham/enumer
//go:generate go run github.com/dmarkham/enumer -linecomment -sql -json -text -yaml -gqlgen -type=ApplicationType,AccessTokenType
````

The first `go:generate` installs the enumer.  Without it, the second `go:generate` would fail unless the enumer was manually installed.

### CODE REVIEW

File: go.mod [hunk 1]
```diff
@@ -15,6 +15,8 @@ require (
 	github.com/sirupsen/logrus v1.8.1
 	github.com/stretchr/testify v1.8.0
 	github.com/zitadel/logging v0.3.3
+	golang.org/x/crypto v0.0.0-20210921155107-089bfa567519 // indirect
+	golang.org/x/net v0.0.0-20211015210444-4f30a5c0130f // indirect
```
muir, line 19: Added by `go mod tidy`

livio-a (approved):

PR #198, opened 2022-07-21
feat: add rp.RefreshAccessToken

muir:
This adds a new function for `RelyingParty`s: `RefreshAccessToken`.

I didn't see any other way to do this.  Did I miss it?   It also makes the `tokenEndpointCaller` interface public so that it's more accessible.

I don't have a test for this.

### DIFF
--- a/pkg/client/client.go
+++ b/pkg/client/client.go
@@ -15,17 +15,20 @@ import (
 	"github.com/zitadel/oidc/pkg/oidc"
 )
 
-var Encoder = func() httphelper.Encoder {
-	e := schema.NewEncoder()
-	e.RegisterEncoder(oidc.SpaceDelimitedArray{}, func(value reflect.Value) string {
-		return value.Interface().(oidc.SpaceDelimitedArray).Encode()
-	})
-	return e
-}()
+var (
+	Encoder = func() httphelper.Encoder {
+		e := schema.NewEncoder()
+		e.RegisterEncoder(oidc.SpaceDelimitedArray{}, func(value reflect.Value) string {
+			return value.Interface().(oidc.SpaceDelimitedArray).Encode()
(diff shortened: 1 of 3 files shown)


livio-a:
@muir I think this fit's perfectly... IMO the `RefreshAccessToken` though should return `oidc.Tokens` (extends oauth.Tokens with the id_token and parsed claims) and use the `VerifyTokens` method to verify and parse the id_token (and access_token)
question then is what happens when no id_token is returned (plain oauth  refresh_token request)
@fforootd do you agree on id_token validation or do you have any opinion on the oauth / oidc topic?

fforootd:
> @muir I think this fit's perfectly... IMO the RefreshAccessToken though should return oidc.Tokens (extends oauth.Tokens with the id_token and parsed claims) and use the VerifyTokens method to verify and parse the id_token (and access_token)
> question then is what happens when no id_token is returned (plain oauth refresh_token request)
> @fforootd do you agree on id_token validation or do you have any opinion on the oauth / oidc topic?

This work fine with me. Thank you both for this!

muir:
@fforootd    I'm not sure what you mean.   Do you want me to take @livio-a 's suggestion?   If so, what should I do if there is no id_token in the reply?  Leave `Tokens.IDToken` empty and set `Tokens.IDTokenClaims` to `nil`?

fforootd:
Sorry for not being specific enough 😁 

Yes I think doing "Leave Tokens.IDToken empty and set Tokens.IDTokenClaims to nil?" is the best way at the moment.

I cant figure a "better" solution currently.

livio-a:
If a client really needs the id_token / claims at that moment i'ts still possible to do the implementation / validation separately after calling `RefreshAccessToken `

PR #202, opened 2022-07-25
chore: add go 1.19 to matrix build

livio-a:


.github/workflows/release.yml
```diff
@@ -15,7 +15,7 @@ jobs:
     runs-on: ubuntu-18.04
```
fforootd, line 15: Can we change this to 20.04?

livio-a: sure... i've updated both

fforootd (approved):

PR #207, opened 2022-08-22
fix: nil pointer dereference on UserInfoAddress

### LINKED ISSUE #203: `invalid memory address or nil pointer dereference` when accessing UserInfoAddress
**Describe the bug**
When user info is missing address field it's decoded into `userinfo` struct where `Address` field has `UserInfoAddress` type but `nil` value.

`UserInfoAddress` type doesn't expect `nil` as receiver, so code like `info.GetAddress().GetCountry()` panics.

**To Reproduce**
1. Unmarshal userinfo without address
2. Try `info.GetAddress().GetCountry()`
3. See panic

**Expected behavior**
Probably, returning default values for all fields (e.g. `""`).

Another (IMO better but breaking) way is to return a pointer. So it will be possible to check value `GetAddress() == nil`

**Additional context**
Temp workaround:
```go
func isNil(val any) bool {
	return val == nil || (reflect.ValueOf(val).Kind() == reflect.Ptr && reflect.ValueOf(val).IsNil())
}

addr := info.GetAddress()
if !isNil(addr) {
  // ...
}
```

muhlemmer:
When the Address field in userinfo is nil, an empty instance of userInfoAddress is now returned. This allows for safe chaining of Get methods. All of an empty userInfoAddress' Get methods will return empty strings.

Fixes #203

### DIFF
--- a/pkg/oidc/userinfo.go
+++ b/pkg/oidc/userinfo.go
@@ -167,6 +167,9 @@ func (u *userinfo) IsPhoneNumberVerified() bool {
 }
 
 func (u *userinfo) GetAddress() UserInfoAddress {
+	if u.Address == nil {
+		return &userInfoAddress{}
+	}
 	return u.Address
 }
 
@@ -389,7 +392,11 @@ func (u *userinfo) UnmarshalJSON(data []byte) error {
 	if err := json.Unmarshal(data, &a); err != nil {
 		return err
 	}
-	u.Address = a.Address
+
+	if a.Address != nil {
+		u.Address = a.Address
+	}
+
 	u.UpdatedAt = Time(time.Unix(a.UpdatedAt, 0).UTC())
 
 	if err := json.Unmarshal(data, &u.claims); err != nil {
--- a/pkg/oidc/userinfo_test.go
+++ b/pkg/oidc/userinfo_test.go
@@ -81,3 +81,39 @@ func TestUserInfoEmailVerifiedUnmarshal(t *testing.T) {
 		}, uie)
 	})
 }
+
+// issue 203 test case.
+func Test_userinfo_GetAddress_issue_203(t *testing.T) {
+	tests := []struct {
+		name string
+		data string
+	}{
+		{
+			name: "with address",
+			data: `{"address":{"street_address":"Test 789\nPostfach 2"},"email":"test","email_verified":true,"name":"Test","phone_number":"0791234567","phone_number_verified":true,"private_claim":"test","sub":"test"}`,
+		},
+		{
+			name: "without address",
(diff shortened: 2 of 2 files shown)


muir (approved):

muhlemmer:
I've updated this PR to catch the nil in `userinfo.UnmarshalJSON` and set a new empty value to the `userinfo.Address` field. The original change in `userinfo.GetAddress()` is reverted. And the tests now use actual json.

pkg/oidc/userinfo.go
```diff
@@ -389,7 +389,13 @@ func (u *userinfo) UnmarshalJSON(data []byte) error {
 	if err := json.Unmarshal(data, &a); err != nil {
 		return err
 	}
-	u.Address = a.Address
+
+	if a.Address == nil {
+		u.Address = new(userInfoAddress)
+	} else {
+		u.Address = a.Address
+	}
```
livio-a, line 397: what about if we only check for nil here:
```go
if a.Address != nil {
	u.Address = a.Address
}
```
and an additional check in the GetAddress():
```go
func (u *userinfo) GetAddress() UserInfoAddress {
	if u.Address == nil {
		return &userInfoAddress{}
	}
	return u.Address
}
```
this way we can avoid the nil pointer and marshalling the userinfo again would return the same values as we had before unmarshalling

pkg/oidc/userinfo_test.go
```diff
@@ -81,3 +81,35 @@ func TestUserInfoEmailVerifiedUnmarshal(t *testing.T) {
...
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			info := &userinfo{}
+			if err := json.Unmarshal([]byte(tt.data), info); err != nil {
+				t.Fatal(err)
+			}
+
+			info.GetAddress().GetCountry() //<- used to panic
```
livio-a, line 112:
```suggestion
			err := json.Unmarshal([]byte(tt.data), info)
			assert.NoError(t, err)

			info.GetAddress().GetCountry() //<- used to panic

			// now shortly assure that a marshalling still produces the same as was parsed into the struct
			marshal, err := json.Marshal(info)
			assert.NoError(t, err)
			assert.Equal(t, tt.data, string(marshal))
```

livio-a (commented):
hey @muhlemmer what do you think about this proposal?

pkg/oidc/userinfo.go
muhlemmer: I agree. Done

muhlemmer:
@livio-a  I've rebased, because the test were failing with my old branch. Comments are resolved.

livio-a (approved):

PR #211, opened 2022-09-16
fix: nil pointer dereference on UserInfoAddress (option 2)

muhlemmer:
This is an alternative to #207, where `userInfoAddress` is no longer accessed through pointers. This makes it always safe for acces. IMHO this is a slightly cleaner approach because we don't need to `if`/`else`.


livio-a:
closed in favour of #207

PR #220, opened 2022-09-28
fix: avoid potential race conditions

muir:
This change avoids five potential race conditions.

Lazy initializes in `RelyingParty` and `OpenIDProvider` operate without locks.  Rather than adding locks, I simply moved the initialization to when the object is created.

The  OpenIDProvider signer updates when there is new data in a channel.  It was doing its work without a lock.  For that one, I added a lock.

### DIFF
--- a/pkg/client/rp/relying_party.go
+++ b/pkg/client/rp/relying_party.go
@@ -190,6 +190,10 @@ func NewRelyingPartyOIDC(issuer, clientID, clientSecret, redirectURI string, sco
 	rp.oauthConfig.Endpoint = endpoints.Endpoint
 	rp.endpoints = endpoints
 
+	// avoid races by calling these early
+	_ = rp.IDTokenVerifier() // sets idTokenVerifier
+	_ = rp.ErrorHandler()    // sets errorHandler
+
 	return rp, nil
 }
 
--- a/pkg/client/rp/relying_party.go
+++ b/pkg/client/rp/relying_party.go
@@ -154,6 +154,8 @@ func NewRelyingPartyOAuth(config *oauth2.Config, options ...Option) (RelyingPart
 		}
 	}
 
+	_ = rp.IDTokenVerifier()
+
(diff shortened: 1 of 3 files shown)


pkg/op/op.go
```diff
@@ -166,6 +166,11 @@ func NewOpenIDProvider(ctx context.Context, config *Config, storage Storage, opO
 
 	o.crypto = NewAESCrypto(config.CryptoKey)
 
+	// Avoid potential race conditions by calling these early
+	_ = o.AccessTokenVerifier()
+	_ = o.JWTProfileVerifier()
+	_ = o.openIDKeySet()
```
livio-a, line 172: wouldn't this also have to be done with for the `IDTokenHintVerifier`as well?

pkg/client/rp/relying_party.go
```diff
@@ -154,6 +154,8 @@ func NewRelyingPartyOAuth(config *oauth2.Config, options ...Option) (RelyingPart
 		}
 	}
 
+	_ = rp.IDTokenVerifier()
```
livio-a, line 157: is it only a problem for the OAuth RP?

pkg/op/op.go
muir: You're quite right.  Thanks!

pkg/client/rp/relying_party.go
muir: I'm not sure I understand the question.   It's a problem if an id token needs to get verified...

livio-a: The "precalls" to the methods or only done in the OAuth provider constructor but not in the OIDC (discovery based) one. If i get this right, this would need them as well:
image

muir: Oh!   Got it.  Thanks.   I'll add that to the PR.   I had been blind to the OAuth constructor.

livio-a (approved):

PR #221, opened 2022-09-28
add access token verifier ops to openidProvider

muir:
For various reasons, I'm not using RS256.   While `RelyingParty` had a way to specify that other `alg`s could be used, OpenIDProivder did not.    This change adds the ability to user other crypto algorithms.

Just as an aside, why not have the default accept more algorithms?   RS384, RS512, ES256, ES384, ES512 at least.

### DIFF
--- a/pkg/op/op.go
+++ b/pkg/op/op.go
@@ -170,20 +170,21 @@ func NewOpenIDProvider(ctx context.Context, config *Config, storage Storage, opO
 }
 
 type openidProvider struct {
-	config              *Config
-	endpoints           *endpoints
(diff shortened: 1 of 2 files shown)


livio-a (approved):

livio-a:
> Just as an aside, why not have the default accept more algorithms? RS384, RS512, ES256, ES384, ES512 at least.

I think the RS256 was just a default because the OIDC spec defines it as must be supported and is default if not provided (e.g. by id_token_signed_response_alg  in the discovery, ...) for id_tokens

PR #224, opened 2022-10-04
chore(example): fix listener usage in app example

mv-kan:


livio-a (approved):
this is a silly mistake... thanks @mv-kan for fixing this

PR #226, opened 2022-10-06
Make supported scopes configurable

wdantuma:
implemented support for configurable supported scopes


livio-a:
Hey @wdantuma 

Thanks for your PR... we'll check this probably tomorrow.

muhlemmer:
Hi, thanks for the pull request. We are currently developing on the `next` branch because we intend to release v2 soon.
As you probably have seen, I've tried to merge this into the `next` branch by resolving merge conflicts. Doing so, I noticed that this PR is in fact changing more than one thing:

1. configurable supported scopes;
2. configurable headers;
3. an additional `dpop` header;

We want to maintain the 1 change per PR philosophy. 

Also, we have changed the way we configure CORS, by dropping the `handler` import, which was one of the merge conflicts.

I apologize that we did not pickup this PR sooner. But as it stands now, this is not merge-able.  If you still need this functionality, I would suggest to create a new PR for each change, based of the `next` branch. I promise I will pick it up quicker ;).

PR #231, opened 2022-10-13
feat: add rp.RevokeToken

muir:
This adds a `RevokeToken()` function to the `rp` package.

Most of the implementation is inside `client`.   It also adjusts the documentation of the `Storage.RevokeToken` to reflect that the first argument can either be a token or a token id and that the second argument, the `userID` is sometimes empty.

`op.Revoke()` will now use status code 500 for server errors.

Lastly, a new markdown file is added that lists things to adjust with the next major release.  Listing these in advance allows implementors to be compatible ahead of time.

These changes are tested by automated tests in a private repo.

### DIFF
--- a/NEXT_RELEASE.md
+++ b/NEXT_RELEASE.md
@@ -1,0 +1,6 @@
+
+# Backwards-incompatible changes to be made in the next major release
+
+- Add `rp/RelyingParty.GetRevokeEndpoint`
+- Rename `op/OpStorage.GetKeyByIDAndUserID` to `op/OpStorage.GetKeyByIDAndClientID` 
+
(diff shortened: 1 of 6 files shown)


muir:
An alternative changing the documentation of `Storage.RevokeToken` would be to add an additional method to `Storage`:

```
// CanRefreshTokenInfo is an optional additional interface that Storage can support.
// Supporting CanRefreshTokenInfo is required to be able to revoke a refresh token that
// does not happen to also be an access token.
type CanRefreshTokenInfo interface {
        GetRefreshTokenInfo(ctx context.Context, clientID string, token string) (userID string, tokenID string, err error)
}
```

livio-a (approved):

PR #238, opened 2022-10-28
chore: add an RP/OP integration test

muir:
This adds an integration test that uses an RP client to create a session with an OP server. 
It refreshes its tokens and then ends the session.

This is built on top of other open pull requests: #229 #230 #231 #237 so this will look like it's changing a lot more than it actually is until those are merged and this is rebased.

Merging this bumps code coverage from 12% to 39%

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -23,7 +23,7 @@ jobs:
         uses: actions/setup-go@v3
         with:
           go-version: ${{ matrix.go }}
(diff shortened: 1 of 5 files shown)


muir:
Once my other remaining pull requests are merged (#244 & #238), I'll rebase on main to resolve the conflicts.

livio-a (commented):
found a single typo... other than that is just really great! thank you very much! 😃

livio-a (approved):

PR #239, opened 2022-11-01
docs: fix zitadel doc uri in the README

Anthony-Jhoiro:
The current uri points to a 404 page, I think that the page name change


fforootd (approved):
Good catch, thanks for the PR!

PR #242, opened 2022-11-11
fix: allow http schema for redirect url for native apps in dev mode

utkuozdemir:
Same that is done on [these lines](https://github.com/zitadel/oidc/blob/main/pkg/op/auth_request.go#L299-L301) needs to be done here as well.


muir (approved):
LGTM, but I don't think my approval matters.

livio-a (approved):

PR #244, opened 2022-11-14
feat: add CanRefreshTokenInfo to support non-JWT refresh tokens 

muir:
Add an additional, optional, op.Storage interface so that refresh tokens
that are not JWTs do not cause failures when they randomly, sometimes, decrypt without error

```go
// CanRefreshTokenInfo is an optional additional interface that Storage can support.
// Supporting CanRefreshTokenInfo is required to be able to revoke a refresh token that
// does not happen to also be a JWTs work properly.
type CanRefreshTokenInfo interface {
        // GetRefreshTokenInfo must return oidc.ErrInvalidRefreshToken when presented
	// with a token that is not a refresh token.
	GetRefreshTokenInfo(ctx context.Context, clientID string, token string) (userID string, tokenID string, err error)
}
```

This will have a minor conflict with #231: both add NEXT_RELEASE.md.  This branch has the version that should be kept.

### DIFF
--- a/pkg/oidc/error.go
+++ b/pkg/oidc/error.go
@@ -18,7 +18,6 @@ const (
 	InteractionRequired  errorType = "interaction_required"
 	LoginRequired        errorType = "login_required"
 	RequestNotSupported  errorType = "request_not_supported"
-	InvalidRefreshToken  errorType = "invalid_refresh_token"
 )
 
 var (
@@ -76,11 +75,6 @@ var (
 	ErrRequestNotSupported = func() *Error {
 		return &Error{
 			ErrorType: RequestNotSupported,
-		}
-	}
-	ErrInvalidRefreshToken = func() *Error {
-		return &Error{
-			ErrorType: InvalidRefreshToken,
 		}
 	}
 )
--- a/pkg/oidc/error.go
+++ b/pkg/oidc/error.go
@@ -18,6 +18,7 @@ const (
 	InteractionRequired  errorType = "interaction_required"
 	LoginRequired        errorType = "login_required"
 	RequestNotSupported  errorType = "request_not_supported"
+	InvalidRefreshToken  errorType = "invalid_refresh_token"
 )
 
 var (
@@ -75,6 +76,11 @@ var (
 	ErrRequestNotSupported = func() *Error {
 		return &Error{
 			ErrorType: RequestNotSupported,
+		}
+	}
+	ErrInvalidRefreshToken = func() *Error {
+		return &Error{
+			ErrorType: InvalidRefreshToken,
 		}
 	}
 )
--- a/pkg/op/storage.go
+++ b/pkg/op/storage.go
@@ -51,8 +51,8 @@ type AuthStorage interface {
 }
 
 // CanRefreshTokenInfo is an optional additional interface that Storage can support.
-// Supporting CanRefreshTokenInfo is required to be able to revoke a refresh token that
-// does not happen to also be JWTs.
+// Supporting CanRefreshTokenInfo is required to be able to (revoke) a refresh token that
+// is neither an encrypted string of <tokenID>:<userID> nor a JWT.
 type CanRefreshTokenInfo interface {
 	// GetRefreshTokenInfo must return oidc.ErrInvalidRefreshToken when presented
 	// with a token that is not a refresh token.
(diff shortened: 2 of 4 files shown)


pkg/oidc/error.go
```diff
@@ -18,6 +18,7 @@ const (
 	InteractionRequired  errorType = "interaction_required"
 	LoginRequired        errorType = "login_required"
 	RequestNotSupported  errorType = "request_not_supported"
+	InvalidRefreshToken  errorType = "invalid_refresh_token"
```
livio-a, line 21: is this error in some RFC? i could only find invalid_grant for this case (https://www.rfc-editor.org/rfc/rfc6749#section-5.2):
> The provided authorization grant (e.g., authorization
               code, resource owner credentials) or refresh token is
               invalid, expired, revoked, does not match the redirection
               URI used in the authorization request, or was issued to
               another client.

pkg/op/storage.go
```diff
@@ -50,6 +50,15 @@ type AuthStorage interface {
 	GetKeySet(context.Context) (*jose.JSONWebKeySet, error)
 }
 
+// CanRefreshTokenInfo is an optional additional interface that Storage can support.
+// Supporting CanRefreshTokenInfo is required to be able to revoke a refresh token that
+// does not happen to also be JWTs.
```
livio-a, line 55: this irritates me...
the zitadel implementation also uses refresh tokens that are not JWTs, but rather opaque strings (as probably most implementations do)
and in fact, if I don't implement the interface, the token will still be decrypted

muir: I'm open to other solutions. 

The problem is in this function:
```go
func getTokenIDAndSubjectForRevocation(ctx context.Context, userinfoProvider UserinfoProvider, accessToken string) (string, string, bool) {
        tokenIDSubject, err := userinfoProvider.Crypto().Decrypt(accessToken)
        if err == nil {
                splitToken := strings.Split(tokenIDSubject, ":")
                if len(splitToken) != 2 {
                        return "", "", false
                }
                return splitToken[0], splitToken[1], true
        }
        accessTokenClaims, err := VerifyAccessToken(ctx, accessToken, userinfoProvider.AccessTokenVerifier())
        if err != nil {
                return "", "", false
        }
        return accessTokenClaims.GetTokenID(), accessTokenClaims.GetSubject(), true
}
```

Prior to my change, it gets called with refresh tokens.  Randomly, some refresh tokens will successfully decrypt.   Randomly, some of the decrypted refresh tokens will have exactly one ":" in them.

pkg/oidc/error.go
muir: No, I made this up.  If that isn't allowed, I'll find another way to signal that the token isn't a refresh token.   I'll just return an extra bool insted.

livio-a: I mean I've found some other implementations (like slack) which return this error to the client, but i couldn't find some spec other than the one mentioned above. Therefore I just want to prevent that this error type is used publicly / as an error in a response. 
So i first thought about having a private error, but you obviously need a public one to be able to return it in the `GetRefreshTokenInfo` implementation.
What do you think about a simple error with a comment? Something like:
```
// ErrInvalidRefreshToken is used internally to state that the provided token is not a refresh token
// it must be returned in `GetRefreshTokenInfo`
var ErrInvalidRefreshToken = errors.New("refresh token is invalid")
```
I'm unsure where to declare this error, here or in storage.go, next to the `CanRefreshTokenInfo` interface

pkg/op/storage.go
livio-a: OK i get the intention. And the current way is naive (<- couldn't think of a better word)
It really makes no sense to just assume that the token is either an opaque token (either access or refresh and both an encrypted form of <tokenID>:<userID>) or a JWT
Maybe we just need to change the description to:
```
// CanRefreshTokenInfo is an optional additional interface that Storage can support.
// Supporting CanRefreshTokenInfo is required to be able to (revoke) a refresh token that
// is neither an encrypted string of <tokenID>:<userID> nor a JWT.
```
Or is this confusing?

muir: It's a little confusing, but it's clear enough.  I've applied that comment.

pkg/oidc/error.go
muir: I've moved the error to op/token_revocation.go.   Would you like it in op/storage.go instead?

muir (commented):
@livio-a I've addressed your comments.

muir:
@livio-a : I've moved the error definition to `op/storage.go`.  Is there anything else holding you back from approving this?

livio-a (approved):

PR #269, opened 2023-01-27
doc: document lack of client caching

muir:
My implementation of `OPStorage` now depends upon behavior that is not currently documented.  This change simply documents that behavior.

Why would I do this?  Well, I want to pass additional information from the `/authorize` request though such that the `LoginURL` is completely different.   I do this by pulling information out of the `http.Request` before `op.Authorize` sees the request.  I add that info to the request `Context` and then use that information when handling the `GetClientByClientID` call so that I can later adjust what `LoginURL` does.   Yes, that's a bit hacky, but it's all good as long as the returned client is never cached.

### DIFF
--- a/pkg/op/storage.go
+++ b/pkg/op/storage.go
@@ -67,6 +67,8 @@ type ClientCredentialsStorage interface {
 }
 
 type OPStorage interface {
+	// GetClientByClientID loads a Client. The returned Client is never cached and is only used to
+	// handle the current request.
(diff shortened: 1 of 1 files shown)


muhlemmer (approved):

muhlemmer:
Thanks for your PR. Can you please rebase against current main, so that the Go 1.20 test is run as part of merge requirements.

muir:
> Thanks for your PR. Can you please rebase against current main, so that the Go 1.20 test is run as part of merge requirements.

done

PR #283, opened 2023-02-17
refactor: use struct types for claim related types

muhlemmer:
BREAKING change.
The following types are changed from interface to struct type:

- AccessTokenClaims
- IDTokenClaims
- IntrospectionResponse
- UserInfo and related types.
- JWTProfileAssertionClaims

The following methods of OPStorage now take a pointer to a struct type,
instead of an interface:

- SetUserinfoFromScopes
- SetUserinfoFromToken
- SetIntrospectionFromToken

The following functions are now generic, so that type-safe extension
of Claims is now possible:

- op.VerifyIDTokenHint
- op.VerifyAccessToken
- rp.VerifyTokens
- rp.VerifyIDToken

This is still a work in process. Need to create some regression tests and I want to verify if the current approach is what we want.


muhlemmer:
Created regression tests, based on `next` branch (4bd80308f63ba0a100ab0dcf09548fd82f875df8) and rebased this PR on that. This already caught some bugs :).

muhlemmer:
**TODO**

Write unit tests for functions:
- [ ] `op.VerifyIDTokenHint`
- [ ] `op.VerifyAccessToken`
- [ ] `rp.VerifyTokens`
- [ ] `rp.VerifyIDToken`
- [ ] Write examples on these functions, how to extend with custom types

Write unit tests for types (getters, setters):
- [x] `AccessTokenClaims`
- [x] `IDTokenClaims`
- [ ] `IntrospectionResponse`
- [ ] `UserInfo` and related types.

pkg/op/verifier_access_token.go
```diff
@@ -18,8 +18,6 @@ type accessTokenVerifier struct {
 	maxAgeIAT         time.Duration
 	offset            time.Duration
 	supportedSignAlgs []string
-	maxAge            time.Duration
-	acr               oidc.ACRVerifier
```
muhlemmer, line 22: wasn't used anywhere

.github/workflows/release.yml
```diff
@@ -16,15 +16,15 @@ jobs:
...
+        go: ['1.18', '1.19', '1.20']
     name: Go ${{ matrix.go }} test
     steps:
       - uses: actions/checkout@v3
       - name: Setup go
         uses: actions/setup-go@v3
         with:
           go-version: ${{ matrix.go }}
-      - run: go test -race -v -coverprofile=profile.cov -coverpkg=github.com/zitadel/oidc/... ./pkg/...
+      - run: go test -race -v -coverprofile=profile.cov -coverpkg=./pkg/...,./example/... ./pkg/...
```
muhlemmer, line 27: perhaps we should skip `./example` here from the coverage report as well. Lots of missed lines and not really relevant.

```suggestion
      - run: go test -race -v -coverprofile=profile.cov -coverpkg=./pkg/... ./pkg/...
```

pkg/op/verifier_access_token_example_test.go
```diff
@@ -0,0 +1,71 @@
...
+	"fmt"
+
+	tu "github.com/zitadel/oidc/v2/internal/testutil"
+	"github.com/zitadel/oidc/v2/pkg/oidc"
+	"github.com/zitadel/oidc/v2/pkg/op"
+)
+
+// MyCustomClaims extends the TokenClaims base,
+// so it implments the oidc.Claims interface.
+// Instead of carying a map, we add needed fields
```
livio-a, line 14:
```suggestion
// so it implements the oidc.Claims interface.
// Instead of carrying a map, we add needed fields
```

pkg/oidc/verifier.go
```diff
@@ -32,6 +32,12 @@ type ClaimsSignature interface {
 	SetSignatureAlgorithm(algorithm jose.SignatureAlgorithm)
 }
 
+type IDClaims interface {
+	Claims
+	GetSignatureAlgorithm() jose.SignatureAlgorithm
```
livio-a, line 37: i think we should either move the `GetSignatureAlgorithm` to the `Claims` interface or implement it on the `IDTokenClaims` struct, because currently the `TokenClaims` implements it, but not the whole `IDClaims` interface, which feels weird:
image

pkg/client/rp/verifier_tokens_example_test.go
```diff
@@ -0,0 +1,87 @@
...
+	"fmt"
+
+	tu "github.com/zitadel/oidc/v2/internal/testutil"
+	"github.com/zitadel/oidc/v2/pkg/client/rp"
+	"github.com/zitadel/oidc/v2/pkg/oidc"
+)
+
+// MyCustomClaims extends the TokenClaims base,
+// so it implments the oidc.Claims interface.
+// Instead of carying a map, we add needed fields
```
livio-a, line 14:
```suggestion
// so it implmeents the oidc.Claims interface.
// Instead of carrying a map, we add needed fields
```

pkg/client/rp/verifier_tokens_example_test.go
```diff
@@ -0,0 +1,87 @@
...
+}
+
+// Nested struct types are also possible.
+type Nested struct {
+	Count int      `json:"count,omitempty"`
+	Tags  []string `json:"tags,omitempty"`
+}
+
+/*
+idToken caries the following claims. foo and bar are custom claims
```
livio-a, line 37:
```suggestion
idToken carries the following claims. foo and bar are custom claims
```

pkg/op/verifier_access_token_example_test.go
```diff
@@ -0,0 +1,71 @@
...
+}
+
+// Nested struct types are also possible.
+type Nested struct {
+	Count int      `json:"count,omitempty"`
+	Tags  []string `json:"tags,omitempty"`
+}
+
+/*
+accessToken caries the following claims. foo and bar are custom claims
```
livio-a, line 34:
```suggestion
accessToken carries the following claims. foo and bar are custom claims
```

pkg/oidc/token_test.go
```diff
@@ -0,0 +1,214 @@
...
+	}
+
+	got := NewAccessTokenClaims(
+		want.Issuer, want.Subject, nil,
+		want.Expiration.AsTime(), want.JWTID, "foo", time.Second,
+	)
+
+	// Make equal not fail on dynamic timestamp
+	got.IssuedAt = 0
+	got.NotBefore = 0
```
livio-a, line 159: would be great if we could test these as well, with range at least

pkg/oidc/token_test.go
```diff
@@ -0,0 +1,214 @@
...
+		want.Issuer, want.Subject, want.Audience,
+		want.Expiration.AsTime(),
+		want.AuthTime.AsTime().Add(time.Second),
+		want.Nonce, want.AuthenticationContextClassReference,
+		want.AuthenticationMethodsReferences, want.AuthorizedParty,
+		time.Second,
+	)
+
+	// Make equal not fail on dynamic timestamp
+	got.IssuedAt = 0
```
livio-a, line 211: same as just above

.github/workflows/release.yml
livio-a: yes i'd go with pkg and remove the example

livio-a (changes requested):

pkg/client/rp/verifier.go
```diff
@@ -21,69 +21,71 @@ type IDTokenVerifier interface {
 
 // VerifyTokens implement the Token Response Validation as defined in OIDC specification
 // https://openid.net/specs/openid-connect-core-1_0.html#TokenResponseValidation
-func VerifyTokens(ctx context.Context, accessToken, idTokenString string, v IDTokenVerifier) (oidc.IDTokenClaims, error) {
-	idToken, err := VerifyIDToken(ctx, idTokenString, v)
+func VerifyTokens[C oidc.IDClaims](ctx context.Context, accessToken, idToken string, v IDTokenVerifier) (claims C, err error) {
+	var nilClaims C
```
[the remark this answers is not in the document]
muhlemmer: correct. Because of type parameters, I can't pass a direct `nil`. For example an implementation of `oidc.IDClaims` might not be a pointer receiver but a struct and a `nil` would be illigal. Hence, Go dis-allows the usage of `nil` for a generic type.

pkg/oidc/verifier.go
muhlemmer: The problem is with `JWTTokenRequest` which does not implement this method.
By setting the method on `Claims`,  In `VerifyJWTAssertion` the following functions will complain:

- `oidc.CheckAudience`
- `oidc.CheckExpiration`
- `oidc.CheckIssuedAt`

`JWTTokenRequest` does have a `SetSignatureAlgorithm` method which is a no-op. I could add `GetSignatureAlgorithm`, but it would always return an empty string. I felt that would be a dirtier solution with the risk that future devs will use `JWTTokenRequest` in places where the difference does matter.

muhlemmer: https://github.com/zitadel/oidc/pull/283#discussion_r1132295959

muhlemmer: image

muhlemmer: Misread your comment, yes I could implement it for `IDTokenClaims` instead, but it already extends the `TokenClaims` type, so it's already implemented by extension. Adding one extra method doesn't hurt tho. Let me know what you want to do.

pkg/oidc/token.go
```diff
@@ -20,380 +18,146 @@ const (
...
+func (c *TokenClaims) GetAudience() []string                          { return c.Audience }
+func (c *TokenClaims) GetExpiration() time.Time                       { return c.Expiration.AsTime() }
+func (c *TokenClaims) GetIssuedAt() time.Time                         { return c.IssuedAt.AsTime() }
+func (c *TokenClaims) GetNonce() string                               { return c.Nonce }
+func (c *TokenClaims) GetAuthTime() time.Time                         { return c.AuthTime.AsTime() }
+func (c *TokenClaims) GetAuthorizedParty() string                     { return c.AuthorizedParty }
+func (c *TokenClaims) GetSignatureAlgorithm() jose.SignatureAlgorithm { return c.SignatureAlg }
+func (c *TokenClaims) GetAuthenticationContextClassReference() string {
+	return c.AuthenticationContextClassReference
 }
```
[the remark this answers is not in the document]
muhlemmer: I've just used the expanded style, I'm good either way,

pkg/oidc/token_test.go
muhlemmer: done, github doesn't mark as outdated because it was solved on the lines above.

livio-a (approved):

PR #285, opened 2023-02-22
feat: implement RFC 8628 - Device authorization grant

### LINKED ISSUE #264: Device authorization grant
As a developer I want to be able to use the device authorization grant for authentication, to be able to authenticate my devices.

**Acceptance Criteria**
- [x] I am able to authenticate with device authorization grant
- [x] Implementation follows the standard https://oauth.net/2/device-flow/
- [x] I am able to authenticate my device
- [x] I am able to use it in the OP
- [x] I am able to use it in the RP

**Additional Context**
Reported issue: #141

muhlemmer:
closes #264


muhlemmer:
Ready for pre-review.

TODO:

- [ ] Unit tests
- [ ] Examples

muhlemmer:
@livio-a  the latest commit includes client authentication for the device endpoint. I've extracted the logic from introspection and made it reusable and tested.

TODO as of now:

- [ ] implement the "code" request param as discussed
- [ ] more Unit tests
- [ ] Examples

pkg/op/device.go
```diff
@@ -0,0 +1,291 @@
...
+		return err
+	}
+
+	response := &oidc.DeviceAuthorizationResponse{
+		DeviceCode:      deviceCode,
+		UserCode:        userCode,
+		VerificationURI: config.UserFormURL,
+	}
+
+	endpoint := o.UserCodeVerificationEndpoint().Absolute(IssuerFromContext(r.Context()))
```
stebenz, line 91: As the parameter name is either a constant or defined as a request parameter for the RP, I would at least define the parameter name not inline

pkg/op/device.go
```diff
@@ -0,0 +1,291 @@
...
+	if err != nil {
+		return err
+	}
+
+	client, err := exchanger.Storage().GetClientByClientID(ctx, clientID)
+	if err != nil {
+		return err
+	}
+	if !clientAuthenticated {
+		if m := client.AuthMethod(); m != oidc.AuthMethodNone { // Livio: Does this mean "public" client?
```
stebenz, line 183: Still some discussion open?

pkg/op/device.go
```diff
@@ -0,0 +1,291 @@
...
+	if err != nil {
+		return err
+	}
+
+	storage, err := assertDeviceStorage(o.Storage())
+	if err != nil {
+		return err
+	}
+
+	ctx := r.Context()
```
stebenz, line 261: not really necessary if the variable is only used once

pkg/op/device.go
```diff
@@ -0,0 +1,291 @@
...
+func ParseUserCodeVerificationRequest(r *http.Request, decoder httphelper.Decoder) (*UserCodeVerificationRequest, error) {
+	if err := r.ParseForm(); err != nil {
+		return nil, oidc.ErrInvalidRequest().WithDescription("cannot parse form").WithParent(err)
+	}
+
+	req := new(UserCodeVerificationRequest)
+	if err := decoder.Decode(req, r.Form); err != nil {
+		return nil, oidc.ErrInvalidRequest().WithDescription("cannot parse user code form").WithParent(err)
+	}
+	if req.Code == "" {
```
stebenz, line 283: What information has to be in "Code" as in the RFC are only user_code and the verification_uri mentioned? If it's not necessary or a secondary attribute from user_code then the request will fail

pkg/op/device_test.go
```diff
@@ -0,0 +1,204 @@
...
+
+func TestNewDeviceCode(t *testing.T) {
+	t.Run("reader error", func(t *testing.T) {
+		runWithRandReader(errReader{}, func() {
+			_, err := NewDeviceCode(16)
+			require.Error(t, err)
+		})
+	})
+
+	t.Run("dirrent lengths, rand reader", func(t *testing.T) {
```
stebenz, line 39: Typo with "dirrent"?

pkg/op/device.go
muhlemmer: @livio-a  suggested I should check if a `Client` is defined as public in storage, if it is not authenticated. I could not find a method called `Public` or similar on the `Client` interface and guessed that the above means the same.

The comment is there so it attracts the attention of the reviewer, so that I can verify my guess is right ;) (I assumed Livio)

stebenz: I just had a talk with Livio, AuthMethodNone doesn't mean public, there would be a function to query if the client is "Confidential" whereas non-confidential means public.
This logic is correct because if you have an authentication method you have to be authenticated, which currently covers every option.

pkg/op/device.go
muhlemmer: as discussed, this handler has been removed

pkg/op/device.go
muhlemmer: as discussed, handler removed

livio-a (commented):
LGTM and tested with the example implementations

i suggest we change the title to `feat: implement RFC 8628 - Device authorization grant` so that ensure to get a new semantic version after the merge

livio-a (approved):

PR #297, opened 2023-03-01
feat: glob support for RedirectURIs

### LINKED ISSUE #293: Wildcard URIs are not valid redirect URIs
**Describe the bug**
Wildcard URIs, like `*` or `https://example.com/*` are not recognized as valid redirect or post-logout redirect URIs.

**To Reproduce**
1. Configure a client with a wildcard redirect URI (for example, `https://example.com/*`)
2. Pass a URI that should satisfy that URI in an authentication request (for example, `https://example.com/test`)
3. See that the URI is not valid

**Expected behavior**
Wildcard URIs should be considered valid, as web applications should be able to have a variable entrypoint, for example if someone were to link to a page that is not the default, or a user is on a page that they want to return to after authenticating

**Desktop (please complete the following information):**
- OS: Debian 10
- Browser N/A

**Additional context**
The problem seems to be [here](https://github.com/zitadel/oidc/blob/main/pkg/op/auth_request.go#L294), for example.  When checking if a URI is valid, it does literal comparison but does not account for wildcards.  This is not the only place where this problem may occur though

muir:
To maintain backwards compatibility, an optional op.Client interface is added:  

```go
// HasRedirectGlobs is an optional interface that can be implemented by implementors of
// Client. See https://pkg.go.dev/github.com/gobwas/glob#Compile for glob
// interpretation. Redirect URIs that match either the non-glob version or the
// glob version will be accepted. Glob URIs are only partially supported for native
// clients: "http://" is not allowed except for loopback or in dev mode.
type HasRedirectGlobs interface {
	RedirectURIGlobs() []string
	PostLogoutRedirectURIGlobs() []string
}
```

In the `next` branch, instead of adding an additional interface, `RedirectURIs` will be changed to return two arrays.

Resolves #293


pkg/op/session.go
```diff
@@ -98,5 +100,16 @@ func ValidateEndSessionPostLogoutRedirectURI(postLogoutRedirectURI string, clien
 			return nil
 		}
 	}
+	if globClient, ok := client.(HasRedirectGlobs); ok {
+		for _, uriGlob := range globClient.PostLogoutRedirectURIGlobs() {
+			matcher, err := glob.Compile(uriGlob)
```
muhlemmer, line 105: Are you aware of https://github.com/gobwas/glob#performance?

I understand that this is the only moment where you can actually compile the globs, as they come from storage and there is no way to store the compiled objects. However, I do would like to see some benchmarks to have an idea about performance impact using this library / approach. It might be just fine, but this sentence triggers me:

> This means, that compilation could take time, but strings matching is done faster,

muhlemmer:
Thanks for working on this! I quickly went over what you've done so far and left a comment. When this PR moves out of draft I will do a more detailed review.

Unless you have a direct need to have this in current stable `main`, I would suggest to go directly for `next`.  Zitadel is already running on a predecessor of `next` for months now and we intend to merge it into `main` within the next 2 weeks. (yes, there is still much refactoring to be done, but probably we just go faster to v3).

pkg/op/session.go
muir: @muhlemmer I could add a LRU cache of compiled globs.  Or perhaps, I should add that to the glob library so that all users benefit?   Adding a cache is pretty easy but adds a dependency and a utility function.

muir:
> Thanks for working on this! I quickly went over what you've done so far and left a comment. When this PR moves out of draft I will do a more detailed review.
> 
> Unless you have a direct need to have this in current stable `main`, I would suggest to go directly for `next`. Zitadel is already running on a predecessor of `next` for months now and we intend to merge it into `main` within the next 2 weeks. (yes, there is still much refactoring to be done, but probably we just go faster to v3).

Hmm.   Do you like my idea for `next` of having RedirectURIs return two slices?

Are there releases of the `next` branch that I can reference?

pkg/op/session.go
muir: The nice LRU cache that is already indirectly part of the oidc package, github.com/hashicorp/golang-lru, has a nice generics-enhanced interface.   Are we ready to use generics in the `next` branch?

muhlemmer:
- Generics: yes. `next` will support generics, and drop any support for <go-1.18 (part of #283, WIP)
- Releases for `next`: there should be a new release coming today and then automatically on each merge: #298
- Regarding 2 slices, lru cache etc I will have think and get back to you later, probably on Discord to have a quick chat.

muhlemmer:
I'm okay with merging this into `main` because it would solve #293 on the short term. However, for `next` I want to investigate if we can find a solution without API or storage change and double slice returns. So don't send a PR for that. TBD in follow-up issue

muhlemmer (approved):

PR #298, opened 2023-03-01
chore: enable github actions for next branch

muhlemmer:


.github/workflows/release.yml
```diff
@@ -3,6 +3,7 @@ on:
   push:
     branches:
       - main
+      - next
```
livio-a, line 6: in the `release` step we check again if it's main
you need to change that one as well and maybe you can also try with `dry_run`

muhlemmer: Done

livio-a (approved):

PR #319, opened 2023-03-09
Test all routes

### LINKED ISSUE #303: implement regression tests for all endpoints in `op`
As developer I want to ensure that changing dependencies for routing, decoding and encoding does not make this framework behave differently towards the outside world. Therefore all endpoints that are currently defined should get a test that validates data input and output through the complete router stack.

## Acceptance criteria
- [x] Each endpoint is called with valid data
- [x] All request fields, including optional, are populated
- [x] Raw return data is stored for later comparation
- [x] Returned and stored data is parsed into objects and compared
- [x] Tests must run in CI/CD.

muhlemmer:
This adds tests for all routes

Closes #303

### DIFF
--- a/example/server/storage/storage.go
+++ b/example/server/storage/storage.go
@@ -859,7 +859,7 @@ func (s *Storage) DenyDeviceAuthorization(ctx context.Context, userCode string) 
 	return nil
 }
 
-// AuthRequestDone is used by testing
+// AuthRequestDone is used by testing and is not required to implement op.Storage
 func (s *Storage) AuthRequestDone(id string) error {
 	s.lock.Lock()
 	defer s.lock.Unlock()
--- a/example/server/storage/storage.go
+++ b/example/server/storage/storage.go
@@ -28,8 +28,8 @@ var serviceKey1 = &rsa.PublicKey{
 	E: 65537,
 }
 
-// var _ op.Storage = &storage{}
-// var _ op.ClientCredentialsStorage = &storage{}
+var _ op.Storage = &Storage{}
+var _ op.ClientCredentialsStorage = &Storage{}
 
 // storage implements the op.Storage interface
 // typically you would implement this as a layer on top of your database
@@ -46,6 +46,7 @@ type Storage struct {
 	signingKey    signingKey
 	deviceCodes   map[string]deviceAuthorizationEntry
 	userCodes     map[string]string
+	serviceUsers  map[string]*Client
 }
 
 type signingKey struct {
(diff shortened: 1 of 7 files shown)


pkg/op/device_test.go
```diff
@@ -16,52 +15,10 @@ import (
...
-
-	var err error
-	testProvider, err = op.NewOpenIDProvider(testIssuer, config,
-		storage.NewStorage(storage.NewUserStore(testIssuer)), op.WithAllowInsecure(),
-	)
-	if err != nil {
-		panic(err)
-	}
-}
-
```
muhlemmer, line 64: note to reviewer: this just got moved to `op_test.go`  as it is now used by the package tests in general, as opposed to just devices.

example/server/storage/storage.go
```diff
@@ -847,3 +858,44 @@ func (s *Storage) DenyDeviceAuthorization(ctx context.Context, userCode string)
 	s.deviceCodes[s.userCodes[userCode]].state.Denied = true
 	return nil
 }
+
+// AuthRequestDone is used by testing
```
muir, line 862:
```suggestion
// AuthRequestDone is used by testing and is not required to implement op.Storage
```

pkg/op/context.go
```diff
@@ -7,16 +7,16 @@ import (
...
 
 type IssuerInterceptor struct {
 	issuerFromRequest IssuerFromRequest
 }
 
-//NewIssuerInterceptor will set the issuer into the context
-//by the provided IssuerFromRequest (e.g. returned from StaticIssuer or IssuerFromHost)
+// NewIssuerInterceptor will set the issuer into the context
+// by the provided IssuerFromRequest (e.g. returned from StaticIssuer or IssuerFromHost)
 func NewIssuerInterceptor(issuerFromRequest IssuerFromRequest) *IssuerInterceptor {
```
muir, line 20: I've written code similar to this in order to pass things through into my implementation of `Storage`.  How about we simply put the `http.Request` into the `context.Context` that is passed into most `Storage` functions.  That would solve this class of problem for most people.  We can also include the decoded body. 

I'm not a fan of Context, but doing this could solve a whole set of situations all at once.

muir (approved):
Nice bump in coverage!

pkg/op/context.go
muhlemmer: The implementation of this function was not part of this PR. Other changes in the file just triggered `gofmt` on the above comments.

If there is additional functionality you want from the context, I suggest to open an issue and we can discuss there.  Note that is you do need a custom context, you can always provide your own middleware that takes care of that:

https://github.com/zitadel/oidc/blob/6f8a392ce20a6063081438b38a43d09bf74d81dd/pkg/op/op.go#L491

stebenz (approved):

PR #320, opened 2023-03-09
fix: unmarshalling of scopes in access token

### LINKED ISSUE #318: `ParseToken` fails parsing an access token due to wrong scope type
**Describe the bug**

When trying to parse an access token using `oidc.ParseToken()` it fails with
```
json: cannot unmarshal string into Go struct field Alias.scope of type []string
```
because the scope of the access token is not an array but a space separated string.

**To Reproduce**
1. Get an access token where scope is something like `"scope":"openid profile email"`
2. Try parsing the access token
```go
claims := oidc.EmptyAccessTokenClaims()

_, err := oidc.ParseToken(accessToken, claims)
if err != nil {
    // it will fail
}
```
3. See error:  `json: cannot unmarshal string into Go struct field Alias.scope of type []string`

**Expected behavior**
It should not fail, but parse the token correctly.

muhlemmer:
The Scopes field in accessTokenClaims should be a  SpaceDelimitedArray, in order to allow for correct unmarshalling.

Fixes #318

### DIFF
--- a/pkg/oidc/token.go
+++ b/pkg/oidc/token.go
@@ -69,23 +69,23 @@ func NewAccessTokenClaims(issuer, subject string, audience []string, expiration 
 }
 
 type accessTokenClaims struct {
-	Issuer                              string   `json:"iss,omitempty"`
-	Subject                             string   `json:"sub,omitempty"`
-	Audience                            Audience `json:"aud,omitempty"`
-	Expiration                          Time     `json:"exp,omitempty"`
-	IssuedAt                            Time     `json:"iat,omitempty"`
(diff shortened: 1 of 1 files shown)


muhlemmer:
@monstermunchkin

Currently we don't have a setter for the `Scopes` field in the library. And therefore also a lack of testing in that manner. Can you please test this fix for me, by pulling this commit into your project:

```
go get github.com/zitadel/oidc@3fb86d10191902b9b07a855ffdacb2cc686d9e46
```

monstermunchkin:
@muhlemmer

This fix works for me, thanks!

livio-a (approved):

PR #328, opened 2023-03-14
chore: document non-standard glob client

### LINKED ISSUE #308: Follow up on Wildcard support
This is a follow issue on, #297 which was a quick-fix for #293 on `main`. However, I would like to see if we can implement `path.Match` without an API and storage change. It would keep usage more simple for the end-user if we can use the same fields for URL input.

I will investigate later this week and come with a proposal.

muhlemmer:
Update the docs and example on glob support regarding standard deviation.
Glob pattern matching was introduced as an optional interface in #297 to provide wildcard support.

1. Fix a typo
2. Document standard deviation when using globs.
3. Add example on how to toggle the underlying client implementation based on DevMode.

Open to discussion.

Closes #308

### DIFF
--- a/pkg/op/client.go
+++ b/pkg/op/client.go
@@ -57,7 +57,7 @@ type Client interface {
 // glob version will be accepted. Glob URIs are only partially supported for native
 // clients: "http://" is not allowed except for loopback or in dev mode.
 //
-// Note that globbing / wildcards are not permitted by the oidc
+// Note that globbing / wildcards are not permitted by the OIDC
 // standard and implementing this interface can have security implications.
 // It is advised to only return a client of this type in rare cases,
 // such as DevMode for the client being enabled.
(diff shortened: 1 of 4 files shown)


muir:
@muhlemmer LGTM.

Other auth servers also support globs.  For example, Keycloak.

pkg/op/client.go
```diff
@@ -56,6 +56,12 @@ type Client interface {
 // interpretation. Redirect URIs that match either the non-glob version or the
 // glob version will be accepted. Glob URIs are only partially supported for native
 // clients: "http://" is not allowed except for loopback or in dev mode.
+//
+// Note that globbing / wildcards are not permitted by the oidc
```
muir, line 60:
```suggestion
// Note that globbing / wildcards are not permitted by the OIDC
```

muir (approved):

stebenz (approved):

PR #345, opened 2023-03-23
feat: add context to client all calls

### LINKED ISSUE #309: client: add context to Call functions
As a developer I would like to have the ability to use a context when calling functions in `/pkg/client/...`. This would allow me to set custom timeouts / deadlines. This is also ground-work to allow the ability to pass loggers along with the context.

Related: #218

## Acceptance criteria

- [x] `http.FormRequest()` to accept a `context.Context` as first argument and setting it in the outgoing `http.Request ;
- [x] All direct and indirect callers of `http.FormRequest()` to accept and pass along a `context.Context` as first argument.

muhlemmer:
BREAKING CHANGE
closes #309

### DIFF
--- a/pkg/client/profile/jwt_profile.go
+++ b/pkg/client/profile/jwt_profile.go
@@ -1,6 +1,7 @@
 package profile
 
 import (
+	"context"
 	"net/http"
 	"time"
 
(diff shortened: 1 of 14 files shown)


pkg/client/profile/jwt_profile.go
```diff
@@ -85,9 +104,13 @@ func (j *jwtProfileTokenSource) HttpClient() *http.Client {
 }
 
 func (j *jwtProfileTokenSource) Token() (*oauth2.Token, error) {
```
stebenz, line 106: Wouldn't a better option be to make the context a parameter here too, so that you don't have to create a context.Background in the function? As this function is used to create a request.

muhlemmer: That was my first intention. However, `Token()` appears to implement the [`oauth2.TokenSource`](https://pkg.go.dev/golang.org/x/oauth2#TokenSource). Therefore I created a local interface that embeds the `TokenSource` and extends it with `TokenCtx()` to allow for calls with context.

https://github.com/zitadel/oidc/blob/6d455501aed493f3d0c8ce3a364ecf1db5f6a306/pkg/client/profile/jwt_profile.go#L15-L18

Another option would be to drop the `oauth2.TokenSource` compatibility, but I'm not sure if that breaks users that specifically need it.

stebenz (approved):

PR #353, opened 2023-03-29
fix: Only set GrantType once

### LINKED ISSUE #352: Device authorization flow fails on polling after first try
**Describe the bug**
When using device authorization flow, polling for the access token fails after the first try with
```
ErrorType=unauthorized_client Description=Grant type 'urn:ietf:params:oauth:grant-type:device_code,urn:ietf:params:oauth:grant-type:device_code' not allowed for the client.
```

That is because `grant_type` contains the `urn:ietf:params:oauth:grant-type:device_code` value twice instead of once. I believe this is caused by the `client.DeviceAccessTokenRequest` struct containing the grant type twice, once in `oidc.ClientCredentialsRequest`, and again in `oidc.DeviceAccessTokenRequest`:

```go
type DeviceAccessTokenRequest struct {
	*oidc.ClientCredentialsRequest
	oidc.DeviceAccessTokenRequest
}
```

This should somehow be deduplicated.

**To Reproduce**
1. Try using the device authorization flow
2. See error: ```ErrorType=unauthorized_client Description=Grant type 'urn:ietf:params:oauth:grant-type:device_code,urn:ietf:params:oauth:grant-type:device_code' not allowed for the client.```

**Expected behavior**
Polling should work correctly, instead of failing straight away.

monstermunchkin:
This fixes an issue where, when using the device authorization flow, the
grant type would be set twice. Some OPs don't accept this, and fail when
polling.

With this fix the grant type is only set once, which will make some OPs
happy again.

Fixes #352

### DIFF
--- a/pkg/client/rp/device.go
+++ b/pkg/client/rp/device.go
@@ -12,7 +12,6 @@ import (
 func newDeviceClientCredentialsRequest(scopes []string, rp RelyingParty) (*oidc.ClientCredentialsRequest, error) {
 	confg := rp.OAuthConfig()
 	req := &oidc.ClientCredentialsRequest{
-		GrantType:    oidc.GrantTypeDeviceCode,
 		Scope:        scopes,
 		ClientID:     confg.ClientID,
 		ClientSecret: confg.ClientSecret,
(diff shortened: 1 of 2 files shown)


muhlemmer (approved):
LGTM, thanks!

PR #356, opened 2023-03-30
feat: Allow modifying request to device authorization endpoint

### LINKED ISSUE #354: Allow setting URL parameters when calling device authorization endpoint
**Is your feature request related to a problem? Please describe.**
When using the device authorization flow, it's currently not possible to set URL parameters when calling the device authorization endpoint. In some cases, the OP would like to receive the `audience` parameter in order to issue a proper access token. 

**Describe the solution you'd like**
It would be nice if URL parameters could somehow be passed when calling [`rp.DeviceAuthorization()`](https://github.com/zitadel/oidc/blob/main/pkg/client/rp/device.go#L35). A solution could be to add an `authFn` argument to both [`rp.DeviceAuthorization()`](https://github.com/zitadel/oidc/blob/main/pkg/client/rp/device.go#L35) and [`client.CallDeviceAuthorizationEndpoint()`](https://github.com/zitadel/oidc/blob/main/pkg/client/client.go#L189) which could then be passed to [`httphelper.FormRequest()`](https://github.com/zitadel/oidc/blob/main/pkg/client/client.go#L190). But this would break backward compatibility.

monstermunchkin:
This change enables the caller to set URL parameters when calling the
device authorization endpoint.

Resolves #354

### DIFF
--- a/example/client/device/device.go
+++ b/example/client/device/device.go
@@ -45,7 +45,7 @@ func main() {
 	}
 
 	logrus.Info("starting device authorization flow")
-	resp, err := rp.DeviceAuthorization(ctx, scopes, provider)
+	resp, err := rp.DeviceAuthorization(ctx, scopes, provider, nil)
 	if err != nil {
 		logrus.Fatal(err)
 	}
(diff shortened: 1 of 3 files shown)


fforootd:
Thanks for the PR, lets us review this soon.

PR #368, opened 2023-04-13
feat: custom endpoint for device authorization

muhlemmer:
This change allows the customization of the `/device_authorization` endpoint.

### DIFF
--- a/pkg/op/op.go
+++ b/pkg/op/op.go
@@ -476,6 +476,16 @@ func WithCustomKeysEndpoint(endpoint Endpoint) Option {
 	}
 }
 
+func WithCustomDeviceAuthorizationEndpoint(endpoint Endpoint) Option {
+	return func(o *Provider) error {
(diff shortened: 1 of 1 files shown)


pkg/op/op.go
```diff
@@ -476,6 +476,16 @@ func WithCustomKeysEndpoint(endpoint Endpoint) Option {
...
+	return func(o *Provider) error {
+		if err := endpoint.Validate(); err != nil {
+			return err
+		}
+		o.endpoints.DeviceAuthorization = endpoint
+		return nil
+	}
+}
+
 func WithCustomEndpoints(auth, token, userInfo, revocation, endSession, keys Endpoint) Option {
```
livio-a, line 489: should we add it here as well?

muhlemmer: That would be a breaking change, mabe for V3?

livio-a (approved):

PR #414, opened 2023-06-22
fix(client/rs): do not error when issuer discovery has no introspection endpoint

### LINKED ISSUE #413: client/rs: error when issuer discovery has no introspection endpoint
**Describe the bug**
When an issuer does not provide an introspection endpoint during discovery, `newResourceServer()` fails with:
```
introspectURL and/or tokenURL is empty: please provide with either `WithStaticEndpoints` or a discovery url
```
Note that a token endpoint is always provided by the issuer.

According to the OpenID specs for [Discovery responses](https://openid.net/specs/openid-connect-discovery-1_0.html#ProviderConfigurationResponse), an introspection endpoint is not mandatory (not defined in the response object) while the token endpoint is required.

Three examples of issuers that do not provide any introspection endpoint are:
* Google: <https://accounts.google.com/.well-known/openid-configuration>
* Spotify: <https://accounts.spotify.com/.well-known/openid-configuration>
* PayPal: <https://www.paypal.com/.well-known/openid-configuration>

This is because of the following (incorrect) check after discovery that assumes that both endpoints must be present:
https://github.com/zitadel/oidc/blob/9e624986aaf6eb8854ffc4702d7d1cc755919ff5/pkg/client/rs/resource_server.go#L83-L85

In reality, the introspection endpoint is valid to be empty.

**To Reproduce**
Consider the following demonstration Go program:
```go
package main

import (
	"fmt"

	"github.com/zitadel/oidc/v2/pkg/client/rs"
)

const (
	Issuer = "https://accounts.google.com"
//	Issuer = "https://accounts.spotify.com"
//	Issuer = "https://www.paypal.com"
	ClientId = "client-id"
	ClientSecret = "client-secret"
)

func main() {
	rscc, err := rs.NewResourceServerClientCredentials(
		Issuer,
		ClientId,
		ClientSecret,
	)
	if err != nil {
		panic(err)
	}

	fmt.Printf("%+v\n", rscc)

	res, err := rs.Introspect(nil, rscc, "test-token")
	if err != nil {
		panic(err)
	}

	fmt.Printf("%+v\n", res)
}
```
When running the above program, an panic is emitted after the error from `NewResourceServerClientCredentials()`:
```
$ go run main.go
panic: introspectURL and/or tokenURL is empty: please provide with either `WithStaticEndpoints` or a discovery url

exit status 2
```

**Expected behavior**
Given that in the demonstration program above I'm using the discovery mechanism, I expect it to not fail when the issuer does not provide an introspection endpoint during discovery as it is an optional feature.

Instead, I expect the `rscc` object to be correctly instantiated and a future call to `rs.Introspect()` fails with an error related to the introspection endpoint not being set. This would allow to still use `rscc` (for the authorizer functionality) but without introspection support.

It can be argued that if an issuer is known to not provide an introspection endpoint, it can be given statically using `WithStaticEndpoints("", "https://some.host/introspection-endpoint")` (or a dummy value) but is highly unlikely that a provider has a manual introspection endpoint and does not provide it in discovery. This would just propagate a confusing error later during `rs.Introspect()`.

**Potential Solution**
I propose to instead accept that discovery may not provide an introspection endpoint, do not fail in this situation, and add a check for an empty introspection endpoint in `rs.Introspect()` to clearly indicate that the issuer does not support introspection.

In addition, discovery should only set `introspectURL` or `tokenURL` if they were not given in static endpoint arguments.

If the above makes sense, I already have this logic implemented here:
https://github.com/hhromic/oidc/compare/main...hhromic:oidc:fix-rs-discovery

As it can be seen, I also made a set of unit tests that cover this issue.
Let me know if you agree with the above proposal to open a PR with these changes.

hhromic:
This PR fixes `client/rs` to not error when an issuer discovery returns no introspection endpoint.

Summary of changes:
* `newResourceServer()` will only fill `rs.tokenURL` and `rs.introspectURL` if they were not given as static endpoints using `WithStaticEndpoints()`. This allows the discovery mechanism to always respect the user-provided static endpoints and only fill them when one or the other was not statically provided. Also, given that `config.introspectURL` is optional in the OIDC Discovery spec, it is no longer enforced to be present after discovery.
* `Introspect()` will now fail if the passed `ResourceServer` does not have an `IntrospectionURL()` set. This allows for graceful errors in case issuers do not provide an introspection endpoint.
* Added unit tests covering all situations using discovery and static endpoints.

Fixes #413 

### Definition of Ready

- [ ] No debug or dead code

### DIFF
--- a/pkg/client/rs/resource_server.go
+++ b/pkg/client/rs/resource_server.go
@@ -77,11 +77,15 @@ func newResourceServer(issuer string, authorizer func() (interface{}, error), op
 		if err != nil {
 			return nil, err
 		}
-		rs.tokenURL = config.TokenEndpoint
-		rs.introspectURL = config.IntrospectionEndpoint
+		if rs.tokenURL == "" {
+			rs.tokenURL = config.TokenEndpoint
+		}
+		if rs.introspectURL == "" {
+			rs.introspectURL = config.IntrospectionEndpoint
+		}
 	}
-	if rs.introspectURL == "" || rs.tokenURL == "" {
-		return nil, errors.New("introspectURL and/or tokenURL is empty: please provide with either `WithStaticEndpoints` or a discovery url")
+	if rs.tokenURL == "" {
+		return nil, errors.New("tokenURL is empty: please provide with either `WithStaticEndpoints` or a discovery url")
 	}
 	rs.authFn = authorizer
 	return rs, nil
@@ -113,6 +117,9 @@ func WithStaticEndpoints(tokenURL, introspectURL string) Option {
 }
 
 func Introspect(ctx context.Context, rp ResourceServer, token string) (*oidc.IntrospectionResponse, error) {
+	if rp.IntrospectionURL() == "" {
+		return nil, errors.New("resource server: introspection URL is empty")
+	}
 	authFn, err := rp.AuthFn()
 	if err != nil {
 		return nil, err
--- a/pkg/client/rs/resource_server_test.go
+++ b/pkg/client/rs/resource_server_test.go
@@ -1,0 +1,219 @@
+package rs
+
+import (
+	"context"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func TestNewResourceServer(t *testing.T) {
+	type args struct {
+		issuer     string
+		authorizer func() (interface{}, error)
+		options    []Option
+	}
+	type wantFields struct {
+		issuer        string
+		tokenURL      string
+		introspectURL string
+		authFn        func() (interface{}, error)
+	}
+	tests := []struct {
+		name       string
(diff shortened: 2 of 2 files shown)


muhlemmer (commented):
Thanks for this excellent contribution and tests!

muhlemmer (approved):

muhlemmer:
@hhromic if you are interested, you can get some contributor swag through this google form: https://docs.google.com/forms/d/e/1FAIpQLSc-8t4G7HHCnxLKdGR7F3vLtKu31f7K4fdlqW4j0cirMYPUEA/viewform

hhromic:
@muhlemmer thanks for the kind offer and very nice to see your company supports open source in such a great way!
For the moment I'm good, this was a very small contribution hehe. Perhaps next time! 🍻

PR #435, opened 2023-08-29
feat: add typ:JWT header to tokens

### LINKED ISSUE #434: Need to add a "typ":"JWT" header to my tokens
### Preflight Checklist

- [X] I could not find a solution in the existing issues, docs, nor discussions
- [X] I have joined the [ZITADEL chat](https://zitadel.com/chat)

### Describe your problem

Due to compatibility issues, I must have a "typ":"JWT" header in the access tokens that I create.
The current APIs do not provide a way to do that.

Ideally, the `Storage` interface would provide some way for me to adjust the Jose Signer options. 

Would it be okay to simply hard-code the type?  Is the op Signer used for anything where that would be a problem?

I'll make a pull request with the following. I'm open to doing something that is more flexible/complicated if this isn't okay.

```
func SignerFromKey(key SigningKey) (jose.Signer, error) {
        signer, err := jose.NewSigner(jose.SigningKey{
                Algorithm: key.SignatureAlgorithm(),
                Key: &jose.JSONWebKey{
                        Key:   key.Key(),
                        KeyID: key.ID(),
                },
        }, (&jose.SignerOptions{}).WithType("JWT"))
        if err != nil {
                return nil, ErrSignerCreationFailed //TODO: log / wrap error?
        }
        return signer, nil
}
```

### Version

_No response_

### Additional Context

I've got a production bug that this will fix so I'll be using my change immediately.

muir:
This tiny change adds a "typ":"JWT" header to the tokens created. 

Alternatively, the Storage interface could be enhanced to allow user control over JWT headers.

Resolves #434

### DIFF
--- a/pkg/op/signer.go
+++ b/pkg/op/signer.go
@@ -6,9 +6,7 @@ import (
 	"gopkg.in/square/go-jose.v2"
 )
 
-var (
-	ErrSignerCreationFailed = errors.New("signer creation failed")
-)
+var ErrSignerCreationFailed = errors.New("signer creation failed")
 
 type SigningKey interface {
 	SignatureAlgorithm() jose.SignatureAlgorithm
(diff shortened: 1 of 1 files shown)


mgiannakopoulos (approved):
:ship: Ship IT!

Issue #440, opened 2023-09-05
proposal(op): new server interface to replace storage


muhlemmer:
This issue proposes a new interface `Server` in the `op` package. It's purpose is to obsolete the current Storage interface and Provider type over time.

### Community feedback

From the community (and at zitadel we feel the same pain):
- Lots op type assertions: https://github.com/zitadel/oidc/discussions/377. Due to the various storage implementation
- The need to customize certain parts, where usually the storage is the limitation.
  - https://github.com/zitadel/oidc/discussions/409
  - https://github.com/zitadel/oidc/issues/393
  - https://github.com/zitadel/oidc/issues/324
  - https://github.com/zitadel/oidc/issues/325
  - https://github.com/zitadel/oidc/issues/326
  - https://github.com/zitadel/oidc/issues/384

### Maintainability

In the current design there is a mix of business logic and the implementation of the OpenID connect / Oauth2 standards. Over time this resulted in overly complex code and duplication of code which is hard to understand and maintain. 

<details>

<summary>Some examples</summary>

https://github.com/zitadel/oidc/blob/daf82a5e041cb1a2279d7e6f8f56d64dd58b4371/pkg/op/token_code.go#L64-L103

https://github.com/zitadel/oidc/blob/daf82a5e041cb1a2279d7e6f8f56d64dd58b4371/pkg/op/token_refresh.go#L89-L129

</details>

### Performance

Due to the business logic in OIDC and its reliance on the Storage interface. We found that there are multiple calls to certain methods. For example, we found for the token endpoint there are cases where `GetUserByID` is called 3 times. If we would sperate the business logic from implementation we could optimized this by a single call. Basically, business logic shouldn't be scope of the OIDC framework.

## Proposed interface

It will be the scope of this framework to:
 - Route requests to their methods;
  - Parse request data into a type that represents the model from the standards;
  - Validate required fields in the request;
  - Provide utility functions that helps implementations build the response;

It will be the scope of the implementer
  - Handle the request data to obtain or modify state in their storage / database;
  - Implement the business logic that is needed to build the response;

```
// Server describes the interface that needs to be implemented to serve
// OpenID Connect and Oauth2 standard requests.
//
// Methods are called after the HTTP route is resolved and
// the request body is parsed into the Request's Data field.
// When a method is called, it can be assumed that required fields,
// as described in their relevant standard, are validated already.
// The Response Data field may be of any type to allow flexibility
// to extend responses with custom fields. There are however requirements
[...]

PR #447, opened 2023-09-21
feat(op): Server interface

muhlemmer:
This PR bring the new Server interface as described in #440 and adds:
- `op.Server` interface;
- `op.webServer`: a HTTP handler taking care of routing and request parsing and validating required request fields.
- `op.LegacyServer`: a wrapper around the current `OpenIDProvider` and `Storage` interfaces.

The interface should be considered experimental.

The following tests have been provided:

- Unit tests for `op.webServer` (near 100%)
- The op router test has been duplicated and runs against the `LegacyServer`.  Some test error outcomes are different, due to the different order of things being validated. (For example client verification is done in a more centralized manner as the old implementation).
- The integration tests run two times, unmodified, against the `op.Provider` and `op.LegacyServer`


example/server/storage/client.go
```diff
@@ -185,7 +185,7 @@ func WebClient(id, secret string, redirectURIs ...string) *Client {
 		authMethod:                     oidc.AuthMethodBasic,
 		loginURL:                       defaultLoginURL,
 		responseTypes:                  []oidc.ResponseType{oidc.ResponseTypeCode},
-		grantTypes:                     []oidc.GrantType{oidc.GrantTypeCode, oidc.GrantTypeRefreshToken},
+		grantTypes:                     oidc.AllGrantTypes,
```
livio-a, line 188: i've thought about this change again... doesn't it mean we have a breaking change (in the client check)?

pkg/op/server.go
```diff
@@ -0,0 +1,346 @@
...
+	http.Redirect(w, r, red.URL, http.StatusFound)
+}
+
+type UnimplementedServer struct{}
+
+// UnimplementedStatusCode is the status code returned for methods
+// that are not yet implemented.
+// Note that this means methods in the sense of the Go interface,
+// and not http methods covered by "501 Not Implemented".
+var UnimplementedStatusCode = http.StatusNotFound
```
livio-a, line 259: i think the 404 works for simple handlers like health / ready but it feels strange if you send wrong data to an endpoint and get a 400 and as soon as everything is compliant you get a 404 because the function is not implemented

livio-a (commented):
i also found some flakiness in the tests

example/server/storage/client.go
muhlemmer: Without this change the following grant types break:

- Device Authorization: was "fixed" on `main` in https://github.com/zitadel/oidc/pull/400, but this is not (yet) included on `next`. 
-  Token Exchange: will indeed have a breaking change. 

The other grant types where already checking the client:

- CodeExchange: https://github.com/zitadel/oidc/blob/f6cb47fbbb4685afec2c3033005a760e6b90b709/pkg/op/token_code.go#L55-L57
- RefreshToken: https://github.com/zitadel/oidc/blob/f6cb47fbbb4685afec2c3033005a760e6b90b709/pkg/op/token_refresh.go#L111-L113
- JWTProfile: doesn't take a client
- ClientCredentials: 
https://github.com/zitadel/oidc/blob/f6cb47fbbb4685afec2c3033005a760e6b90b709/pkg/op/token_client_credentials.go#L93-L95

livio-a (approved):

PR #484, opened 2023-11-14
feat: Allow CORS policy to be configured

korylprince:
This allows the CORS policy to be configured or disabled on an `op.Provider` via `WithCORSOptions` or on `on.RegisterServer` via `WithServerCORSOptions`. The reason for this feature is library users may want a stricter CORS policy than the default policy currently forced by the library.

This takes care not to break backwards compatibility with any interfaces, functions, or default usage (e.g. the default policy is still applied by default).

Also, this removes the CORS middleware from the `intercept` function as every upstream use of it already had the CORS middleware applied.

### DIFF
--- a/pkg/op/server_http.go
+++ b/pkg/op/server_http.go
@@ -38,8 +38,9 @@ func RegisterServer(server Server, endpoints Endpoints, options ...ServerOption)
 	}
 
 	ws.createRouter()
+	ws.handler = ws.router
 	if ws.corsOpts != nil {
-		return cors.New(*ws.corsOpts).Handler(ws)
+		ws.handler = cors.New(*ws.corsOpts).Handler(ws.router)
 	}
 	return ws
 }
@@ -88,6 +89,7 @@ func WithFallbackLogger(logger *slog.Logger) ServerOption {
 type webServer struct {
 	server    Server
 	router    *chi.Mux
+	handler   http.Handler
 	endpoints Endpoints
 	decoder   httphelper.Decoder
 	corsOpts  *cors.Options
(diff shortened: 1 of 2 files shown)


radsec:
@korylprince ...Requesting if this can also be back-ported to the `v2` implementation as well.

korylprince:
@radsec PR #485 is the backported version of this.

muhlemmer (dismissed):
Thanks! looks great

pkg/op/server_http.go
```diff
@@ -29,15 +29,18 @@ func RegisterServer(server Server, endpoints Endpoints, options ...ServerOption)
...
-	ws.router.Use(cors.New(defaultCORSOptions).Handler)
 
 	for _, option := range options {
 		option(ws)
 	}
 
 	ws.createRouter()
+	if ws.corsOpts != nil {
+		return cors.New(*ws.corsOpts).Handler(ws)
+	}
```
muhlemmer, line 43: Tests currently panic on this, as the test does type assertion to check the server state.  I tried moving the function registration to the router, only then chi panics because it wants middleware before defining routes. And currently the `LegacyServer`  defines a custom route as part of ServerOptions.

The only way I see now to add CORS to the Server is to use the `WithHTTPMiddleware` directly. I don't have time now to get my head around it, so I'll get back to it later.

korylprince:
@muhlemmer thanks for taking a look at this! I pushed a new commit that I think works around this issue nicely. Instead of wrapping the *webServer itself for CORS, keep an internal handler that can either be the router or the CORS-wrapped router, and use that in ServeHTTP.

muhlemmer (approved):
Perfect, thanks!

muhlemmer:
hey @korylprince  thanks for the contribution.

PR #495, opened 2023-12-04
feat(op): User-configurable claims_supported

snowkat:
This change adds a new `SupportedClaims` field to `op.Config` along with `(op.Configuration).SupportedClaims()`, in order to applications to define what claims they support in discovery.

To retain compatibility with the existing system, `(op.Config).SupportedClaims` can be nil, in which case `(op.Config).SupportedClaims()` returns `DefaultSupportedClaims`&mdash;which corresponds to the hardcoded values previously returned.

`op.SupportedClaims(c Configuration)` exists for compatibility as well&mdash;since it's exported, I don't want to assume it's unused by some application&mdash;but can be deprecated.

### Definition of Ready

- [X] I am happy with the code
- [X] Short description of the feature/issue is added in the pr description
- [X] No debug or dead code
- [X] My code has no repetitions
- [X] Critical parts are tested automatically
- [X] Where possible E2E tests are implemented

### DIFF
--- a/pkg/op/config.go
+++ b/pkg/op/config.go
@@ -47,7 +47,6 @@ type Configuration interface {
 	RequestObjectSupported() bool
 	RequestObjectSigningAlgorithmsSupported() []string
 
-	SupportedClaims() []string
 	SupportedUILocales() []language.Tag
 	DeviceAuthorization() DeviceAuthorizationConfig
 }
--- a/pkg/op/config.go
+++ b/pkg/op/config.go
@@ -47,6 +47,7 @@ type Configuration interface {
 	RequestObjectSupported() bool
 	RequestObjectSigningAlgorithmsSupported() []string
 
+	SupportedClaims() []string
 	SupportedUILocales() []language.Tag
 	DeviceAuthorization() DeviceAuthorizationConfig
 }
--- a/pkg/op/discovery.go
+++ b/pkg/op/discovery.go
@@ -57,7 +57,7 @@ func CreateDiscoveryConfig(ctx context.Context, config Configuration, storage Di
 		IntrospectionEndpointAuthMethodsSupported:          AuthMethodsIntrospectionEndpoint(config),
 		RevocationEndpointAuthSigningAlgValuesSupported:    RevocationSigAlgorithms(config),
 		RevocationEndpointAuthMethodsSupported:             AuthMethodsRevocationEndpoint(config),
-		ClaimsSupported:                                    SupportedClaims(config),
+		ClaimsSupported:                                    config.SupportedClaims(),
 		CodeChallengeMethodsSupported:                      CodeChallengeMethods(config),
 		UILocalesSupported:                                 config.SupportedUILocales(),
 		RequestParameterSupported:                          config.RequestObjectSupported(),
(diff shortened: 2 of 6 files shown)


pkg/op/config.go
```diff
@@ -47,6 +47,7 @@ type Configuration interface {
 	RequestObjectSupported() bool
 	RequestObjectSigningAlgorithmsSupported() []string
 
+	SupportedClaims() []string
```
muhlemmer, line 50: Adding a method to the interface would be a breaking change, instead we should use the global function.

pkg/op/discovery.go
```diff
@@ -57,7 +57,7 @@ func CreateDiscoveryConfig(ctx context.Context, config Configuration, storage Di
 		IntrospectionEndpointAuthMethodsSupported:          AuthMethodsIntrospectionEndpoint(config),
 		RevocationEndpointAuthSigningAlgValuesSupported:    RevocationSigAlgorithms(config),
 		RevocationEndpointAuthMethodsSupported:             AuthMethodsRevocationEndpoint(config),
-		ClaimsSupported:                                    SupportedClaims(config),
+		ClaimsSupported:                                    config.SupportedClaims(),
 		CodeChallengeMethodsSupported:                      CodeChallengeMethods(config),
```
muhlemmer, line 61:
```suggestion
		ClaimsSupported:                                    SupportedClaims(config),
```

pkg/op/discovery.go
```diff
@@ -88,7 +88,7 @@ func createDiscoveryConfigV2(ctx context.Context, config Configuration, storage
...
```
muhlemmer, line 92:
```suggestion
		ClaimsSupported:                                    SupportedClaims(config),
```

pkg/op/discovery.go
```diff
@@ -213,32 +213,7 @@ func AuthMethodsRevocationEndpoint(c Configuration) []oidc.AuthMethod {
...
-		"family_name",
-		"given_name",
-		"locale",
-		"email",
-		"email_verified",
-		"phone_number",
-		"phone_number_verified",
-	}
+	return c.SupportedClaims()
 }
```
muhlemmer, line 217:
```suggestion
func SupportedClaims(c Configuration) []string {
	provider, ok := c.(*Provider)
	if ok && provider.config.SupportedClaims != nil {
		return provider.config.SupportedClaims
	}

	return DefaultSupportedClaims
}
```

pkg/op/op.go
```diff
@@ -386,6 +414,14 @@ func (o *Provider) RequestObjectSigningAlgorithmsSupported() []string {
 	return []string{"RS256"}
 }
 
+func (o *Provider) SupportedClaims() []string {
+	if o.config.SupportedClaims == nil {
+		return DefaultSupportedClaims
+	} else {
+		return o.config.SupportedClaims
+	}
+}
```
muhlemmer, line 423: Remove unneeded method

muhlemmer (changes requested):
Thanks for the contribution!

We prefer to avoid breaking changes in the interfaces unless we are aiming for a new major release.

I let some comments on how to prevent this breaking change.

snowkat:
Thanks for the suggestion. I've updated the interface and `op.SupportedClaims`, and reverted my change to the relevant test.

muhlemmer (approved):

PR #503, opened 2023-12-19
feat(rp): Add UnauthorizedHandler

jkroepke:
Any callback error which results into an unauthorized status just print as plain text to the user.

I would like to enrich the user experience and offer an HTML styled error message.

My first idea was to re-use the existing `errorHandler`, but it always returns `http.StatusInternalServerError`. The best possible solution would be an errorHandler which has the preferred http status as function input. However this breaks the function signature.

### Definition of Ready

- [ ] No debug or dead code

### DIFF
--- a/pkg/client/rp/relying_party.go
+++ b/pkg/client/rp/relying_party.go
@@ -9,7 +9,7 @@ import (
 	"net/url"
 	"time"
 
-	jose "github.com/go-jose/go-jose/v3"
+	"github.com/go-jose/go-jose/v3"
 	"github.com/google/uuid"
 	"github.com/zitadel/logging"
 	"golang.org/x/exp/slog"
@@ -67,18 +67,25 @@ type RelyingParty interface {
 
 	// IDTokenVerifier returns the verifier used for oidc id_token verification
 	IDTokenVerifier() *IDTokenVerifier
-	// ErrorHandler returns the handler used for callback errors
 
+	// ErrorHandler returns the handler used for callback errors
 	ErrorHandler() func(http.ResponseWriter, *http.Request, string, string, string)
+
+	// UnauthorizedHandler returns the handler used for unauthorized errors
+	UnauthorizedHandler() func(http.ResponseWriter, *http.Request, string, string)
 
 	// Logger from the context, or a fallback if set.
 	Logger(context.Context) (logger *slog.Logger, ok bool)
 }
 
 type ErrorHandler func(w http.ResponseWriter, r *http.Request, errorType string, errorDesc string, state string)
+type UnauthorizedHandler func(w http.ResponseWriter, r *http.Request, desc string, state string)
 
 var DefaultErrorHandler ErrorHandler = func(w http.ResponseWriter, r *http.Request, errorType string, errorDesc string, state string) {
 	http.Error(w, errorType+": "+errorDesc, http.StatusInternalServerError)
+}
+var DefaultUnauthorizedHandler UnauthorizedHandler = func(w http.ResponseWriter, r *http.Request, desc string, state string) {
+	http.Error(w, desc, http.StatusUnauthorized)
 }
 
 type relyingParty struct {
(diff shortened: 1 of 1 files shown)


pkg/client/rp/relying_party.go
```diff
@@ -67,19 +67,26 @@ type RelyingParty interface {
...
 	// IDTokenVerifier returns the verifier used for oidc id_token verification
 	IDTokenVerifier() *IDTokenVerifier
-	// ErrorHandler returns the handler used for callback errors
 
+	// ErrorHandler returns the handler used for callback errors
 	ErrorHandler() func(http.ResponseWriter, *http.Request, string, string, string)
 
+	// UnauthorizedHandler returns the handler used for unauthorized errors
+	UnauthorizedHandler() func(http.ResponseWriter, *http.Request, string, string)
+
```
muhlemmer, line 76: Adding a method to the interface would be a breaking change. For the proposed implementation we don't need to expose the method, right?

pkg/client/rp/relying_party.go
```diff
@@ -157,6 +165,13 @@ func (rp *relyingParty) ErrorHandler() func(http.ResponseWriter, *http.Request,
 	return rp.errorHandler
 }
 
+func (rp *relyingParty) UnauthorizedHandler() func(http.ResponseWriter, *http.Request, string, string) {
+	if rp.unauthorizedHandler == nil {
+		rp.unauthorizedHandler = DefaultUnauthorizedHandler
+	}
+	return rp.unauthorizedHandler
```
muhlemmer, line 172: race condition on the RP. Better to handle the population of the `rp.unauthorizedHandler` once in the constructor.

muhlemmer (changes requested):

pkg/client/rp/relying_party.go
jkroepke: Hm, I do copy and paste from existing function. The race conditions goes to the both functions, too?

https://github.com/zitadel/oidc/blob/2b35eeb83560d69c6b492eb5dcad02ad30c6da55/pkg/client/rp/relying_party.go#L146-L158

pkg/client/rp/relying_party.go
jkroepke: Correct, we do not need to expose the method. however CodeExchangeHandler is using the interface as input. Without declare the function on the interface, the method is not available.

https://github.com/zitadel/oidc/blob/2b35eeb83560d69c6b492eb5dcad02ad30c6da55/pkg/client/rp/relying_party.go#L448

As I mention, I would prefer to re-use the existing `rp.ErrorHandler`, but it does not fit in this case here.

jkroepke:
@muhlemmer do you an alternative solution here? If I had the choice, I would think about an breaking change which modifies the function signature of the existing ErrorHandler by accepting the http status as input parameter and use the ErrorHandler on all places. I think this is more convenient as having two functions.

muhlemmer:
For this PR to the main branch you could use an optional interface:

```go
type HasUnauthorizedHandler interface {
	// UnauthorizedHandler returns the handler used for unauthorized errors
	UnauthorizedHandler() UnauthorizedHandler
}
```

Then use type-assertion in a helper function to use either the new or old behavior:

```go
func unauthorizedError(w http.ResponseWriter, r *http.Request, desc string, state string, rp RelyingParty) {
	if rp, ok := rp.(HasUnauthorizedHandler); ok {
		rp.UnauthorizedHandler()(w, r, desc, state)
		return
	}
	http.Error(w, desc, http.StatusUnauthorized)
}
```

And call that function where required:

```go
if err := trySetStateCookie(w, state, rp); err != nil {
    unauthorizedError(w, r, "failed to create state cookie: "+err.Error(), state, rp)
    return
}
```

After we merged this we can create a new PR against the `next` branch and do the breaking changes for the v4 release.

jkroepke:
Sounds interesting, I was not aware of pattern. Thanks for bringing this in.

muhlemmer (approved):

PR #571, opened 2024-03-14
feat: return oidc.Error in case of call token failure

celian-garcia:
Hello, Azure API is returning a well formatted oauth error when we try to generate a token with client credentials grant type for example. Respecting that RFC https://www.rfc-editor.org/rfc/rfc6749#section-5.2

But this library doesn't return a proper oidc.Error in that case which is too bad. So I took inspiration from the ``CallDeviceAccessTokenEndpoint`` and I propose you this PR.

### Definition of Ready

- [ ] No debug or dead code

### DIFF
--- a/pkg/client/client.go
+++ b/pkg/client/client.go
@@ -2,7 +2,6 @@ package client
 
 import (
 	"context"
-	"encoding/json"
 	"errors"
 	"fmt"
 	"io"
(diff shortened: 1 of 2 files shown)


pkg/client/client.go
```diff
@@ -71,10 +71,27 @@ func callTokenEndpoint(ctx context.Context, request any, authFn any, caller Toke
...
+	resp := new(struct {
+		*oidc.AccessTokenResponse
+		*oidc.Error
+	})
+	if err = json.NewDecoder(httpResp.Body).Decode(resp); err != nil {
+		return nil, err
+	}
+
+	if httpResp.StatusCode < http.StatusOK || httpResp.StatusCode >= http.StatusBadRequest {
+		return nil, resp.Error
```
muhlemmer, line 90: If for some reason there was an error code and the OP did not return a proper JSON formatted OIDC error, there is a change that resp.Error will remain `nil` here.

muhlemmer (commented):
I agree completely that we should unmarshal and return the oidc.Error. But I would propose to implement it differently.

In `http.HttpRequest`:

https://github.com/zitadel/oidc/blob/56397f88d5b8abcda55ead200af66093e428786e/pkg/http/http.go#L68-L70

1. If the status is not OK, 200, try to unmarshal the response to oidc.Error.
2. If unmarshal returned an error, or if oidc.Error is empty after unmarshal, return an error with the body as string (current solution)
3. Callers, like `callTokenEndpoint`, can use [error wrapping](https://pkg.go.dev/errors), so further down the line the oidc.Error can still be obtained through [`errors.As`](https://pkg.go.dev/errors#As()).

This way we have a central way of handling the oidc.Error and all endpoints may benefit from the solution.

celian-garcia:
@muhlemmer thanks for the review and very detailed answer! 
I was not sure that httprequest was not used in a different context than oidc/oauth calls. I will do the changes you listed :+1: 

Isn't it up to the users of the lib to use ``errors.As`` anyway? I don't see a place where I need to use it inside the ``callTokenEndpoint``. In device auth yes, I can modify the ``CallDeviceAccessTokenEndpoint`` to use then httprequest returning ``oidc.Error`` and use ``errors.As`` in the Poll helper. Maybe this is what you meant.

muhlemmer:
Yes, you are right. `errors.As` should be used by the external callers.

muhlemmer (approved):

Issue #635, opened 2024-08-15
refresh token with Okta fails


muir:
`rp` client calls to Okta to refresh tokens fail.

The error is `The client_assertion_type is invalid.`

What is being sent to Okta is an empty string:

```
client_assertion=&client_assertion_type=&client_id= ....
```

When adding a Singer, the client assertion isn't valid, but Okta complains that the audience is incorrect.

If that is fixed then Okta complains about a missing JWK. I haven't fixed this last one as fixing the first turns out to be easy.

PR #599, opened 2024-04-30
fix: Omit non-standard, empty fields in `RefreshTokenRequest` when performing a token refresh

ymarcus93:
The OIDC spec's definition of a refresh request does not include `client_assertion` or `client_assertion_type` as valid parameters for the refresh request. See request format here: https://openid.net/specs/openid-connect-core-1_0.html#RefreshingAccessToken. The document only displays `client_id`, `client_secret`, `grant_type`, `refresh_token`, and `scope` as acceptable parameters.

Therefore, I propose we add the `omitempty` tags to the `ClientAssertion` and `ClientAssertionType` fields in `RefreshTokenRequest`, so that the token refresh functionality provided by `rp.RefreshTokens` can work with identity providers that may have additional logic or different expectations when these fields are included in the refresh token request.

For example, when attempting to construct an OIDC client via `rp.RelyingParty` against an Okta identity provider, I ran into issues when performing refresh with `rp.RefreshTokens`. The Okta identity provider returned `http status not ok: 400 Bad Request {"error":"invalid_request","error_description":"The client_assertion_type is invalid."}` as an error. I assume I'm receiving this error because I'm calling the `RefreshTokens` func with `clientAssertion=""` and `clientAssertionType=""`; the addition of the `omitempty` tags resolves this issue and hopefully future proofs this library against other identity providers that have the same behavior.

### Definition of Ready

- [X] I am happy with the code
- [X] Short description of the feature/issue is added in the pr description
- [ ] PR is linked to the corresponding user story
- [ ] Acceptance criteria are met
- [ ] All open todos and follow ups are defined in a new ticket and justified
- [ ] Deviations from the acceptance criteria and design are agreed with the PO and documented.
- [ ] No debug or dead code
- [ ] My code has no repetitions
- [ ] Critical parts are tested automatically
- [ ] Where possible E2E tests are implemented
- [ ] Documentation/examples are up-to-date
- [ ] All non-functional requirements are met
- [ ] Functionality of the acceptance criteria is checked manually on the dev system.

PR #636, opened 2024-08-15
fix: client assertions for Okta

muir:
This PR makes two changes to address 
https://github.com/zitadel/oidc/issues/635

It adds `omitempty` on the `client_assertion` for token refresh requests. This allows requests without assertions to work with Okta.

It also adjusts the audience in `client_assertions` so that Okta doesn't complain.

I haven't succeeded with a client assertion with Okta as I haven't yet figured out what I need to with respect to JWKS and Okta.


pkg/oidc/token_request.go
```diff
@@ -74,8 +74,8 @@ type AccessTokenRequest struct {
 	ClientID            string `schema:"client_id"`
 	ClientSecret        string `schema:"client_secret"`
 	CodeVerifier        string `schema:"code_verifier"`
-	ClientAssertion     string `schema:"client_assertion"`
-	ClientAssertionType string `schema:"client_assertion_type"`
+	ClientAssertion     string `schema:"client_assertion,omitempty"`
+	ClientAssertionType string `schema:"client_assertion_type,omitempty"`
```
muhlemmer, line 78: For consistency we should add `omitempty` to all optional fields, like the `client_secret` and `code_verifier`. There might be a chance OP like okta would fail as well, if one would send an assertion and the secret will be an empty value in that case.

muhlemmer (commented):
@muir, I'm confused. You mentioned refresh tokens in the issue, but your fix addresses the CodeExchangeHandler. How do you call the Refresh token? Because `rp.RefreshTokens` uses the `rp.RefreshTokenRequest` type for the request schema. That already has `omitempty` tags:

https://github.com/zitadel/oidc/blob/de034c8d24498884e82adf7027059dec80aafd95/pkg/client/rp/relying_party.go#L735-L743

I still think it is a good idea to add them to `oidc.TokenRequest`, but see my comment on the diff.

muir:
> @muir, I'm confused. You mentioned refresh tokens in the issue, but your fix addresses the CodeExchangeHandler. How do you call the Refresh token? Because `rp.RefreshTokens` uses the `rp.RefreshTokenRequest` type for the request schema. That already has `omitempty` tags:
> 
> https://github.com/zitadel/oidc/blob/de034c8d24498884e82adf7027059dec80aafd95/pkg/client/rp/relying_party.go#L735-L743
> 
> I still think it is a good idea to add them to `oidc.TokenRequest`, but see my comment on the diff.

I'm confused too. I am calling `RefreshTokens`. My code:

```
token, err := rp.RefreshTokens[*oidc.IDTokenClaims](ctx, m.RelyingParty, string(session.RefreshToken), "", "")
```

Figured it out. Testing this is hard-ish and I didn't test with a clean 3.27.0 release. I put in my change before testing.

I think what I needed to do was upgrade. The fix was made three months ago by @ymarcus93 in #599 and my change is redundant. Perhaps still useful but no longer required to fix refresh with Okta.

pkg/oidc/token_request.go
muhlemmer: @muir I'm not seeing it. Did you forget to commit or push?

muir: Hmm, I'm not sure what happened. Pushing, again.

muhlemmer (approved):

PR #653, opened 2024-09-16
feat(oidc): return defined error when discovery failed

ay4toh5i:
This PR adds a specific error to handle discovery failures. 
It allows proper error handling when the issuer URL is invalid or the configuration is missing.
Thank you.

### Definition of Ready
- [x] No debug or dead code
- [x] My code has no repetitions

### DIFF
--- a/pkg/client/client_test.go
+++ b/pkg/client/client_test.go
@@ -46,12 +46,7 @@ func TestDiscover(t *testing.T) {
 	for _, tt := range tests {
 		t.Run(tt.name, func(t *testing.T) {
 			got, err := Discover(context.Background(), tt.args.issuer, http.DefaultClient, tt.args.wellKnownUrl...)
-			if tt.wantErr != nil {
-				assert.ErrorIs(t, err, tt.wantErr)
-				return
-			}
-
-			require.NoError(t, err)
+			require.ErrorIs(t, err, tt.wantErr)
 			if tt.wantFields == nil {
 				return
 			}
--- a/pkg/client/client_test.go
+++ b/pkg/client/client_test.go
@@ -40,9 +40,6 @@ func TestDiscover(t *testing.T) {
 			args: args{
 				issuer: "https://example.com",
 			},
-			wantFields: &wantFields{
-				UILocalesSupported: true,
-			},
 			wantErr: oidc.ErrDiscoveryFailed,
 		},
 	}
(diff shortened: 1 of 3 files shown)


muhlemmer (changes requested):

pkg/client/client_test.go
```diff
@@ -32,16 +33,27 @@ func TestDiscover(t *testing.T) {
...
 			got, err := Discover(context.Background(), tt.args.issuer, http.DefaultClient, tt.args.wellKnownUrl...)
-			if tt.wantErr {
-				assert.Error(t, err)
+			if tt.wantErr != nil {
+				assert.ErrorIs(t, err, tt.wantErr)
 				return
 			}
+
 			require.NoError(t, err)
 			if tt.wantFields == nil {
```
muhlemmer, line 58: Error check be simplefied:

```suggestion
			got, err := Discover(context.Background(), tt.args.issuer, http.DefaultClient, tt.args.wellKnownUrl...)
			require.ErrorIs(t, err, tt.wantErr)
```

pkg/client/client_test.go
```diff
@@ -32,16 +33,27 @@ func TestDiscover(t *testing.T) {
...
 		t.Run(tt.name, func(t *testing.T) {
 			got, err := Discover(context.Background(), tt.args.issuer, http.DefaultClient, tt.args.wellKnownUrl...)
-			if tt.wantErr {
-				assert.Error(t, err)
+			if tt.wantErr != nil {
+				assert.ErrorIs(t, err, tt.wantErr)
 				return
 			}
+
 			require.NoError(t, err)
```
muhlemmer, line 57: Error check can be simplefied:

```suggestion
            require.ErrorIs(t, err, tt.wantErr)
```

ay4toh5i: I understand the suggestion to simplify the error check, but the current implementation uses early returns when an error is encountered to skip the remaining tests. If we simplify the check as proposed, it may cause the tests to fail.

ay4toh5i:
@muhlemmer 
Thanks for taking the time to review my PR and for the feedback! 
I’ve addressed the points you mentioned. Could you take another look when you get a chance?

pkg/client/client_test.go
muhlemmer: When the expectation in `wantFields` is correct (`nil` in this case) the test will not fail.

ay4toh5i: It seems I initially misunderstood your suggestion, but after reconsidering, I’ve incorporated your proposed changes. Thanks!

PR #656, opened 2024-09-24
feat(example): Make it possible to load user configuration from file

lanseg:
This change allows to load users from the json file. 

That should help to create better test scenarios when using example server as fake Zitadel instance for testing. For example, having several users with different permissions or presence in the internal base (known user, unknown user, blocked user, etc).

We don't need dynamic server for testing, but  I could update it to make env variables unified if needed.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -27,7 +27,7 @@ jobs:
         with:
           go-version: ${{ matrix.go }}
       - run: go test -race -v -coverprofile=profile.cov -coverpkg=./pkg/... ./pkg/...
-      - uses: codecov/codecov-action@v4.5.0
+      - uses: codecov/codecov-action@v4.6.0
         with:
           file: ./profile.cov
           name: codecov-go


example/server/users.json
muhlemmer, line 1: This file doesn't need to be included, right? The defaults are already hard-coded in the program.
If you do think it is useful to include the file as an example, then use the [ `embed`](https://pkg.go.dev/embed) package and remove the hard-coded data from the `storage/user.go` file.

muhlemmer (changes requested):
Left 1 comment.

Also please change the PR description to match semver: `feat(example): <description>`

example/server/users.json
lanseg: That is the testing data, I decided to include it as a sample for end users. If that is against some conventions, it's totally ok to remove it.

muhlemmer: It's duplicate data, the convention is called [DRY](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself). So please remove it or `embed` it.

lanseg: I'm very sorry for closing this branch. I wanted to pull latest version before submitting a proper pull request, but removed my own code. 
I'll try to submit a new one soon.

PR #663, opened 2024-10-09
feat(example): Allow configuring some parameters with env variables

lanseg:
This change allows to load users from the json file.

That should help to create better test scenarios when using example server as fake Zitadel instance for testing. For example, having several users with different permissions or presence in the internal base (known user, unknown user, blocked user, etc).

We don't need dynamic server for testing, but I could update it to make env variables unified if needed.

This is an extended version of another pull request: https://github.com/zitadel/oidc/pull/656, which was deleted by mistake.

### DIFF
--- a/example/server/config/config.go
+++ b/example/server/config/config.go
@@ -1,0 +1,40 @@
+package config
+
+import (
+	"os"
+	"strings"
+)
+
+const (
+	// default port for the http server to run
+	DefaultIssuerPort = "9998"
+)
+
+type Config struct {
+	Port        string
+	RedirectURI []string
+	UsersFile   string
+}
+
+// FromEnvVars loads configuration parameters from environment variables.
+// If there is no such variable defined, then use default values.
+func FromEnvVars(defaults *Config) *Config {
+	if defaults == nil {
+		defaults = &Config{}
+	}
+	cfg := &Config{
+		Port:        defaults.Port,
+		RedirectURI: defaults.RedirectURI,
+		UsersFile:   defaults.UsersFile,
+	}
+	if value, ok := os.LookupEnv("PORT"); ok {
+		cfg.Port = value
+	}
+	if value, ok := os.LookupEnv("USERS_FILE"); ok {
+		cfg.UsersFile = value
+	}
+	if value, ok := os.LookupEnv("REDIRECT_URI"); ok {
+		cfg.RedirectURI = strings.Split(value, ",")
+	}
+	return cfg
+}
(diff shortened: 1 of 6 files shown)


example/server/config/config.go
```diff
@@ -0,0 +1,45 @@
...
+	RedirectURIs []string
+	UsersFile    string
+	Issuer       string
+}
+
+// FromEnvVars loads configuration parameters from environment variables.
+// If there is no such variable defined, then use default values.
+func FromEnvVars(defaults *Config) *Config {
+	if defaults == nil {
+		defaults = &Config{}
```
muhlemmer, line 24: Shouldn't `DefaultIssuerPort` be set to the Config here?

example/server/config/config.go
```diff
@@ -0,0 +1,45 @@
...
+		UsersFile:    defaults.UsersFile,
+		Issuer:       defaults.Issuer,
+	}
+	if value, ok := os.LookupEnv("PORT"); ok {
+		cfg.Port = value
+	}
+	if value, ok := os.LookupEnv("USERS_FILE"); ok {
+		cfg.UsersFile = value
+	}
+	if value, ok := os.LookupEnv("REDIRECT_URIS"); ok {
```
muhlemmer, line 38: Please keep the environment variable name `REDIRECT_URI`, so we don't break existing uses of this variables.
```suggestion
	if value, ok := os.LookupEnv("REDIRECT_URI"); ok {
```

muhlemmer (changes requested):

muhlemmer:
Hi @lanseg I noticed you pushed another commit with additional changes after the initial review was done, not related to the review comments. Although I do not have a problem with the additional change, this is not a nice netiquette. If you intent to continue work on a PR, it should be marked as "draft" and not "ready for review". Please keep that in mind for future contributions to any github project.

You also have introduced a build error that needs to be fixed.

I'll mark this as draft for the moment and please mark it ready for review once you are done.

example/server/config/config.go
lanseg: That works as planned - if the defaults is nil, then user didn't want to have any defaults and expects everything to be set in the env vars.

lanseg:
> Hi @lanseg I noticed you pushed another commit with additional changes after the initial review was done, not related to the review comments. Although I do not have a problem with the additional change, this is not a nice netiquette. If you intent to continue work on a PR, it should be marked as "draft" and not "ready for review". Please keep that in mind for future contributions to any github project.
> 
> You also have introduced a build error that needs to be fixed.
> 
> I'll mark this as draft for the moment and please mark it ready for review once you are done.

I'm very sorry for causing a mess, github workflow confused me a bit better.

muhlemmer (approved):

Issue #660, opened 2024-10-04
Access token response is not compliant with OAuth 2.0 (RFC 6749)


isegura-eos-eng:
### Preflight Checklist
* [x]  I could not find a solution in the existing issues, docs, nor discussions
* [ ]  I have joined the [ZITADEL chat](https://zitadel.com/chat)

### Describe your problem
In the RFC 6749 document describing the [Access Token](https://datatracker.ietf.org/doc/html/rfc6749#section-3.3) `scope` request parameter it indicates the following:

> If the issued access token scope is different from the one requested by the client, the authorization server MUST include the "scope" response parameter to inform the client of the actual scope granted.

I've implemented both `authorization_code` and `client_credentials` in my program with the `op.Storage` interface. However, when I supply the `scope` parameter in the token endpoint the response does not include the required parameter.

I think I don't need to supply my code because it's easy to see that `oidc.AccessTokenResponse` does not include the `scope` parameter: https://github.com/zitadel/oidc/blob/97d7b28fc080f9929e590ef1989880b0762e0098/pkg/oidc/token.go#L232

### Describe your ideal solution
Be compliant with OAuth 2.0 in this case. Whenever the `scope` parameter is sent to the token endpoint, check whether the returned scope matches the request. Or return the scope always if that is easier, which complies also with the rule.

### Version
3.30.0

### Environment
Self-hosted

### Additional Context
I'm not sure if I this filed this issue correctly. In the OpenID spec, in the [token endpoint](https://openid.net/specs/openid-connect-core-1_0.html#TokenEndpoint) section indicates to follow "Section 3.2" of RFC 6749. Clearly, 3.3 is not mentioned. Should 3.3 be included?

Aside of this, I think is very helpful to obtain the actual scope given in the access token response. I may have missed if there is other way to achieve this, so please forgive me and educate me if this is the case.

muhlemmer:
Found it: https://github.com/zitadel/zitadel/issues/6609

At the time someone volunteered for a fix, but went silent. @isegura-eos-eng do you want to send a PR for this?

isegura-eos-eng:
[...]

PR #664, opened 2024-10-09
fix(op): add scope to access token scope

isegura-eos-eng:
This attempts to solve [Issue 660](https://github.com/zitadel/oidc/issues/660).

All the details are there. TLDR; Access Token response is not compliant with OAuth 2.0. We decided to _always_ return the `scope` property in the Access Token response to comply with [RFC 6749](https://datatracker.ietf.org/doc/html/rfc6749#section-3.3).

The implementation consists of getting the scope from the token request, which should be already containing only the scopes that are granted.

### Definition of Ready

- [ ] No debug or dead code

### Further improvements

In my opinion, a more elegant solution would be to strictly adhere with the protocol and only return the scope if has been modified from the request. If compliant, clients could assume all scopes are given by the absence of the `scope` property in the token response, this would avoid parsing the response scope. However, this change would probably cause a breaking change in the exported package, so this solution is good for now.

### DIFF
--- a/pkg/op/server_http_routes_test.go
+++ b/pkg/op/server_http_routes_test.go
@@ -145,7 +145,7 @@ func TestServerRoutes(t *testing.T) {
 				"assertion":  jwtProfileToken,
 			},
 			wantCode: http.StatusOK,
-			contains: []string{`{"access_token":`, `"token_type":"Bearer","expires_in":299}`},
+			contains: []string{`{"access_token":`, `"token_type":"Bearer","expires_in":299,"scope":"openid"}`},
 		},
 		{
 			name:      "Token exchange",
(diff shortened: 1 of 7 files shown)


pkg/op/server_http_routes_test.go
```diff
@@ -145,7 +145,7 @@ func TestServerRoutes(t *testing.T) {
 				"assertion":  jwtProfileToken,
 			},
 			wantCode: http.StatusOK,
-			contains: []string{`{"access_token":`, `"token_type":"Bearer","expires_in":299}`},
+			contains: []string{`{"access_token":`, `"token_type":"Bearer","expires_in":299,"scope":"openid"}`},
```
isegura-eos-eng, line 148: I'd like a second opinion on this. It's hard for me to trace back and figure out why this test passes without the `offline_access` scope in the response.

muhlemmer: Hmm but it's weird it does work for client credentials.

isegura-eos-eng: Any clue at why this is happening?

muhlemmer: The example OP storage has the JWT scope validation hard-coded to allow only `openid` as a scope and others are stripped.:

https://github.com/zitadel/oidc/blob/87ab01115708a2a05c20cfd26db82cc7e0e8e338/example/server/storage/storage.go#L566-L576

The Client Credentials does not have a similar step.

For me it is fine to keep it like it is now. So if you're ready you can mark the PR for review.

muhlemmer (approved):

muhlemmer:
hey @isegura-eos-eng thanks for the contribution.

isegura-eos-eng:
Thanks @muhlemmer ! Will do!

Issue #672, opened 2024-10-31
[Bug]: Invalid locale prevents userinfo lookup


muir:
### Preflight Checklist

- [x] I could not find a solution in the documentation, the existing issues or discussions
- [x] I have joined the [ZITADEL chat](https://zitadel.com/chat)

### Version

v3.27.0

### Describe the problem caused by this bug

The `userinfo` call is failing due to an marshal problem with the locale. I don't care about locale. At all. I do care that a paying customer cannot log in.

### To reproduce

I'm not sure what value is being presented by Okta. Their IT says that all their users have locale=null but somehow some can log in and others cannot.

### Expected behavior

Ignore locale if it's not avalid

### Additional Context

I'll propose a fix to ignore all unmarshal errors with locale.

muhlemmer:
> I'm not sure what value is being presented by Okta. Their IT says that all their users have locale=null but somehow some can log in and others cannot.

Yet, it would be really good to know what they _are_ sending. JSON `null` works. Don't tell me those muppets are sending a quoted `"null"` string.

On a side note, for your future debugging efforts, we have added the [Error.WithReturnParentToClient](https://pkg.go.dev/github.com/zitadel/oidc/v3/pkg/oidc#Error.WithReturnParentToClient) which can help understanding these type of errors.

PR #673, opened 2024-10-31
fix: ignore all unmarshal errors from locale

muir:
This changes unmarshal of Locale so that rather than ignoring just `language.ValueError` errors, it ignores all errors. Generally speaking, a bad locale value should not prevent users from authenticating.

Resolves  https://github.com/zitadel/oidc/issues/672

### DIFF
--- a/pkg/oidc/types.go
+++ b/pkg/oidc/types.go
@@ -3,7 +3,6 @@ package oidc
 import (
 	"database/sql/driver"
 	"encoding/json"
-	"errors"
 	"fmt"
 	"reflect"
 	"strings"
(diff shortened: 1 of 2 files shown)


muhlemmer (approved):

muhlemmer:
I'm approving this fix as I understand the rationale that locale is not so important as failed authentications. However, I have a dislike of ignoring errors in general, as it may hamper debug efforts. If this ever backfires in complaints where people couldn't figure out why the Locale is missing (silent error) we might rollback this change.

I do recommend pressuring Okta to do their job and fix this on their side for the sake of long-term stability.

muir:
@muhlemmer We could preserve the raw value sent somewhere in the object and allow it to be interrogated.  Would you like me to do that?

muhlemmer:
The raw values should already be available in the [`Userinfo.Claims`](https://pkg.go.dev/github.com/zitadel/oidc/v3@v3.32.1/pkg/oidc#UserInfo) map. So no need to do that.

muir:
> The raw values should already be available in the [`Userinfo.Claims`](https://pkg.go.dev/github.com/zitadel/oidc/v3@v3.32.1/pkg/oidc#UserInfo) map. So no need to do that.

@muhlemmer 
Hmm, in that case I think the value may be `""`

muhlemmer:
Good one. Yes, with an empty string on the previous version I can reproduce. I was under the impression `omitempty` on the Userinfo field should catch that. :confused:

I will open a seperate PR which reverts and tries to catch the empty string. Perhaps you can try that out instead, and if it works we can go that way.

PR #678, opened 2024-11-11
fix: ignore empty json strings for locale

muhlemmer:
This change reverts fbf009fe75dac732dde39e0eb6fe324b337675e0 so that any non-value error is returned again.
Instead, a check is done for an empty JSON string.

Extra test cases of missing / null / empty values are also added.

This is an alternative solution for  https://github.com/zitadel/oidc/issues/672 

### Definition of Ready

- [x] Acceptance criteria are met
- [x] All open todos and follow ups are defined in a new ticket and justified
- [x] Deviations from the acceptance criteria and design are agreed with the PO and documented.
- [x] No debug or dead code
- [x] My code has no repetitions
- [x] Critical parts are tested automatically
- [x] Where possible E2E tests are implemented
- [x] Documentation/examples are up-to-date
- [x] All non-functional requirements are met
- [x] Functionality of the acceptance criteria is checked manually on the dev system.


muhlemmer:
@muir can you have a look if this alternative does not re-introduce the okta issue you had?

muir:
> @muir can you have a look if this alternative does not re-introduce the okta issue you had?

This should be fine. I don't think I would ever want authorization to be blocked by a bad local though -- I'm not using the locale and thus it has no value.

livio-a (approved):

PR #691, opened 2024-12-17
fix: allow native applications to use https:// on loopback redirect addresses

conblem:
This pull request addresses an issue where native clients were unable to utilize HTTPS for redirects on loopback addresses.

**Changes Made:**
- Updated the redirect validation logic to accommodate HTTPS uri's for native applications

Part of this: https://github.com/zitadel/zitadel/issues/4091

### DIFF
--- a/pkg/op/auth_request.go
+++ b/pkg/op/auth_request.go
@@ -312,11 +312,11 @@ func ValidateAuthReqRedirectURI(client Client, uri string, responseType oidc.Res
 		return oidc.ErrInvalidRequestRedirectURI().WithDescription("The redirect_uri is missing in the request. " +
 			"Please ensure it is added to the request. If you have any questions, you may contact the administrator of the application.")
 	}
-	if strings.HasPrefix(uri, "https://") {
-		return checkURIAgainstRedirects(client, uri)
-	}
 	if client.ApplicationType() == ApplicationTypeNative {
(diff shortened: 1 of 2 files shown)


pkg/op/auth_request.go
```diff
@@ -338,12 +338,15 @@ func ValidateAuthReqRedirectURI(client Client, uri string, responseType oidc.Res
...
 func validateAuthReqRedirectURINative(client Client, uri string) error {
 	parsedURL, isLoopback := HTTPLoopbackOrLocalhost(uri)
-	isCustomSchema := !strings.HasPrefix(uri, "http://")
+	isCustomSchema := !(strings.HasPrefix(uri, "http://") || strings.HasPrefix(uri, "https://"))
 	if err := checkURIAgainstRedirects(client, uri); err == nil {
 		if client.DevMode() {
 			return nil
 		}
-		// The RedirectURIs are only valid for native clients when localhost or non-"http://"
+		if !isLoopback && strings.HasPrefix(uri, "https://") {
```
conblem, line 346: To still support URI's like https://registered/callback this condition had to be added. Im not 100% happy with the code tho

muhlemmer: Why do we need to support `https://registered/callback` on a non-loopback address? Native clients only run on the localhost, right? I also do not see that use case in the linked issue.

muhlemmer (approved):

PR #720, opened 2025-02-26
fix: verify state if cookie exists

vogtp:
If the state is both in the cookie and the request, validate that they are the same.

### Definition of Ready

- [ x] I am happy with the code
- [ x] Short description of the feature/issue is added in the pr description
- [ x] No debug or dead code

With help of https://github.com/Elyes-Ferjani

### DIFF
--- a/pkg/client/rp/relying_party.go
+++ b/pkg/client/rp/relying_party.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"encoding/base64"
 	"errors"
+	"fmt"
 	"log/slog"
 	"net/http"
 	"net/url"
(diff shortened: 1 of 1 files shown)


vogtp:
https://datatracker.ietf.org/doc/html/rfc6749#section-10.12

* What's the point of having a state and not verifying it?
* Returning the state from the form if there is no cookie is a strange choice, but I kept it for compatibility reasons.  BUT IMHO it opens up the lib for replay attacks.

muhlemmer:
I wasn't trying to open a discussion on the importance of the PR. But we require a certain level of explanation in the PR description, so that when we merge it we have a good commit log. So, can you please change the [PR title](https://github.com/zitadel/oidc/blob/main/CONTRIBUTING.md#make-a-pr) to follow semantic release and amend the description describing the problem that is being solved.

vogtp:
This PR holds no more importance to me, since I moved to the coreos oidc implementation by now.
Main reason was I found an other bug in the nonce validation where you compare the generator funcs instead of the values.
The coreos implementation is not certified, but the code seems more mature (at least I do not find bugs whenever I look into the libs) and the way it integrates with the std go libs seems more goish.

Issue #471, opened 2023-10-30
PKCE support is not enough


ay4toh5i:
### Preflight Checklist

- [X] I could not find a solution in the existing issues, docs, nor discussions
- [x] I have joined the [ZITADEL chat](https://zitadel.com/chat)

### Describe your problem

Thanks such a nice library!

I notice that pkce is verified only when the auth method is none in code exchenge.
https://github.com/zitadel/oidc/blob/d58ab6a11584cf458cde09643588b2abeb7c5221/pkg/op/token_code.go#L96-L103

However, pkce should be verified in other methods to mitigate vulnerabilities as decribed below.

> PKCE is not a form of client authentication, and PKCE is not a replacement for a client secret or other [client authentication](https://oauth.net/2/client-authentication/). PKCE is recommended even if a client is using a client secret or other form of client authentication like private_key_jwt.

https://oauth.net/2/pkce/

### Describe your ideal solution

If pkce params requested in authorize, verifying that as possible is ideal.

### Environment

Self-hosted

muhlemmer:
Although I agree with the recommendation, oauth.net is not a definitive resource for the standard, which is described in RFCs. What you describe is Oauth V2.1 I believe. We already have an issue open for that: https://github.com/zitadel/oidc/issues/254. You can vote there with a thumbs up and we can assign priorities based on that.

At the moment we are compliant with Oauth V2. Closing as duplicate.

Issue #254, opened 2022-12-06
OAuth 2.1 compliance regarding PKCE and client secret


stebenz:
**Is your feature request related to a problem? Please describe.**
The current definition of OAuth 2.1 expects a client secret when a confidential app is used in combination with PKCE.

**Describe the solution you'd like**
Check of client secret if PKCE flow is used when app is typed as confidential.

**Describe alternatives you've considered**
Split of compliance with a different version, to bet at least different from the new version of OAuth.

**Additional context**

https://datatracker.ietf.org/doc/html/draft-ietf-oauth-v2-1-09#authorization_codes

ay4toh5i:
PR: #721

I’ve updated the code to verify PKCE even when the auth method isn’t set to “none”. What do you think? I’d really appreciate your feedback and consideration.

PR #721, opened 2025-03-01
feat(op): always verify code challenge when available

ay4toh5i:
refs: #254, #471

Finally the RFC `Best Current Practice for OAuth 2.0 Security` has been approved.

According to the RFC:

> Authorization servers MUST support PKCE [[RFC7636](https://datatracker.ietf.org/doc/html/rfc7636)].
> 
> If a client sends a valid PKCE code_challenge parameter in the authorization request, the authorization server MUST enforce the correct usage of code_verifier at the token endpoint.

Isn’t it time we strengthen PKCE support a bit more?

This PR updates the logic so that PKCE is always verified, even when the Auth Method is not "none".
Could you please review this PR?

And please let me know if there’s anything missing or if there’s anything else I should address.

### Definition of Ready

- [ ] No debug or dead code


example/client/app/app.go
```diff
@@ -33,6 +34,10 @@ func main() {
 	port := os.Getenv("PORT")
 	scopes := strings.Split(os.Getenv("SCOPES"), " ")
 	responseMode := os.Getenv("RESPONSE_MODE")
+	pkce, err := strconv.ParseBool(os.Getenv("PKCE"))
+	if err != nil {
+		logrus.Fatalf("error parsing PKCE %s", err.Error())
+	}
```
muhlemmer, line 41: When `PKCE` is not set, the example fails with:

```
FATA[0000] error parsing PKCE strconv.ParseBool: parsing "": invalid syntax 
exit status 1
```

The unset variant needs to be handled first. 

```suggestion

	var pkce bool
	if pkceEnv, ok := os.LookupEnv("PKCE"); ok {
		var err error
		pkce, err = strconv.ParseBool(pkceEnv)
		if err != nil {
			logrus.Fatalf("error parsing PKCE %s", err.Error())
		}
	}
```

pkg/op/token_code.go
```diff
@@ -84,6 +84,16 @@ func AuthorizeCodeClient(ctx context.Context, tokenReq *oidc.AccessTokenRequest,
 			return nil, nil, err
 		}
 		request, err = AuthRequestByCode(ctx, exchanger.Storage(), tokenReq.Code)
+
+		codeChallenge := request.GetCodeChallenge()
+		if codeChallenge != nil && codeChallenge.Challenge != "" {
+			err = AuthorizeCodeChallenge(tokenReq.CodeVerifier, request.GetCodeChallenge())
+
```
muhlemmer, line 91: Pass the codeChallenge variable here instead, no need to re-get.

pkg/op/token_code.go
```diff
@@ -109,6 +119,16 @@ func AuthorizeCodeClient(ctx context.Context, tokenReq *oidc.AccessTokenRequest,
...
 	request, err = AuthRequestByCode(ctx, exchanger.Storage(), tokenReq.Code)
+
+	codeChallenge := request.GetCodeChallenge()
+	if codeChallenge != nil && codeChallenge.Challenge != "" {
+		err = AuthorizeCodeChallenge(tokenReq.CodeVerifier, request.GetCodeChallenge())
+
+		if err != nil {
+			return nil, nil, err
+		}
+	}
```
muhlemmer, line 130: Second, redundant check?

pkg/op/token_code.go
```diff
@@ -84,6 +84,16 @@ func AuthorizeCodeClient(ctx context.Context, tokenReq *oidc.AccessTokenRequest,
 			return nil, nil, err
 		}
 		request, err = AuthRequestByCode(ctx, exchanger.Storage(), tokenReq.Code)
+
+		codeChallenge := request.GetCodeChallenge()
+		if codeChallenge != nil && codeChallenge.Challenge != "" {
```
muhlemmer, line 89: I don't think we should check for an empty string here.

muhlemmer (commented):
With this change `AuthorizeCodeChallenge` is called 3 different times within `AuthorizeCodeClient` depending on the code path.

Can we simplify the function to only do a single check, perhaps in a higher scope at the start of the `AuthorizeCodeClient`  function?  

See how we do it in Zitadel for example:

https://github.com/zitadel/zitadel/blob/306ce97828aefa567e2bc0b1d207f98a65cf7ba7/internal/api/oidc/token_code.go#L64-L68

muhlemmer (commented):
Can you please add to the PR description what is being implemented or fixed?

- Is there an active bug that describes something is broken if this check is not done?
- Is this an enhancement recomended by a standard? Please link to the the section of the standard.

muhlemmer:
Oops. responded to the wrong PR with the [last comment](https://github.com/zitadel/oidc/pull/721#pullrequestreview-2678054073). Disregard that please.

example/client/app/app.go
ay4toh5i: I overlooked that! Thanks for pointing it out—I’ve incorporated your suggestion.

Issue #405, opened 2023-06-06
The current `rp.VerifyIDToken`'s `azp` check is too strict and doesn't works with OPs like google


wzy9607:
**Is your feature request related to a problem? Please describe.**

https://github.com/zitadel/oidc/blob/e8262cbf1fc224e8a7447cfc3d1ffa93c1a70b3c/pkg/client/rp/verifier.go#L49-L53

Currently, `VerifyIDToken` checks `aud` and `azp` against the same ClientID. However, id token issued by some OP, e.g. Google, can have two different client_id in `aud` and `azp` claims.

When using Google identity on Android, the `azp` claim is the Android app's client_id, while the `aud` claim is the web application's client_id, as [Google's document](https://developers.google.com/identity/openid-connect/openid-connect#an-id-tokens-payload) and [this discussion](https://stackoverflow.com/a/41240814) and [this oidc spec issue](https://bitbucket.org/openid/connect/issues/973/). This means that id token issued by google can't pass `VerifyIDToken`.

**Describe the solution you'd like**

Make `azp` claim check configurable and allow user pass-in a function in `NewIDTokenVerifier` to specify how they want to check the claim.
e.g. an option `func WithAuthorizedPartyChecker(func (azp string, aud []string) error) VerifierOption`

**Describe alternatives you've considered**

Add an option to pass-in a list of `client_id` when `NewIDTokenVerifier`, and check `azp` against that list.

**Additional context**

The OIDC work group seems to have changed the `azp` validation in an errata, see <https://bitbucket.org/openid/connect/src/b84078b1aeb694a79823f3de5a22315df700b22f/openid-connect-core-1_0.xml#lines-1933:1944>, making it optional. But for some reason they haven't updated their website to include the errata yet.

muhlemmer:
Thanks for pointing it out. We will discuss. CC @livio-a

muhlemmer:
Checked the standards and `azp` has indeed become an extension outside the scope of OIDC. We can remove the check and allow an option to be passed that enables a custom check, as @wzy9607 proposed.

Putting this to the backlog, but  a PR is also welcome.

On a side note, reading the google docs you linked:

> The client_id of the authorized presenter. This claim is only needed when the party requesting the ID token is not the same as the audience of the ID token. This may be the case at Google for hybrid apps where a web application and Android app have a different OAuth 2.0 client_id but share the same Google APIs project.

[...]

PR #752, opened 2025-05-28
Make strict Authorized Party check in `rp.VerifyIDToken` optional

Joerger:
This PR makes the default Authorized Party check in `rp.VerifyIDToken` optional by adding an options parameter for dynamic verification functions. This check is meant to be an optional validation requirement, so some providers (including GCP) do not adhere to it.

See https://github.com/zitadel/oidc/issues/405 for more context.

Closes https://github.com/zitadel/oidc/issues/405

### DIFF
--- a/pkg/client/rp/verifier.go
+++ b/pkg/client/rp/verifier.go
@@ -10,15 +10,18 @@ import (
 	"github.com/zitadel/oidc/v3/pkg/oidc"
 )
 
+// VerifyTokenFn is a custom verification function for use in [VerifyIDToken]
+type VerifyTokenFn func(claims oidc.Claims, v *IDTokenVerifier) error
+
 // VerifyTokens implement the Token Response Validation as defined in OIDC specification
 // https://openid.net/specs/openid-connect-core-1_0.html#TokenResponseValidation
-func VerifyTokens[C oidc.IDClaims](ctx context.Context, accessToken, idToken string, v *IDTokenVerifier) (claims C, err error) {
+func VerifyTokens[C oidc.IDClaims](ctx context.Context, accessToken, idToken string, v *IDTokenVerifier, options ...VerifyTokenFn) (claims C, err error) {
 	ctx, span := client.Tracer.Start(ctx, "VerifyTokens")
 	defer span.End()
 
 	var nilClaims C
 
-	claims, err = VerifyIDToken[C](ctx, idToken, v)
+	claims, err = VerifyIDToken[C](ctx, idToken, v, options...)
 	if err != nil {
 		return nilClaims, err
 	}
@@ -30,7 +33,7 @@ func VerifyTokens[C oidc.IDClaims](ctx context.Context, accessToken, idToken str
 
 // VerifyIDToken validates the id token according to
 // https://openid.net/specs/openid-connect-core-1_0.html#IDTokenValidation
-func VerifyIDToken[C oidc.Claims](ctx context.Context, token string, v *IDTokenVerifier) (claims C, err error) {
+func VerifyIDToken[C oidc.Claims](ctx context.Context, token string, v *IDTokenVerifier, options ...VerifyTokenFn) (claims C, err error) {
 	ctx, span := client.Tracer.Start(ctx, "VerifyIDToken")
 	defer span.End()
 
@@ -54,10 +57,6 @@ func VerifyIDToken[C oidc.Claims](ctx context.Context, token string, v *IDTokenV
 	}
 
 	if err = oidc.CheckAudience(claims, v.ClientID); err != nil {
-		return nilClaims, err
-	}
-
-	if err = oidc.CheckAuthorizedParty(claims, v.ClientID); err != nil {
 		return nilClaims, err
 	}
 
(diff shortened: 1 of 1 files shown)


muhlemmer (commented):
This is a slight breaking change. Making the AZP check optional and disabled by default. This allows a security risk being introduced into system which rely on the claim being checked already.

Also note that `VerifyTokens` is also [called by the `RelyinParty`](https://github.com/joerger/oidc/blob/c5c849a93c9fa0a07c6e03202b94b54baa5e804d/pkg/client/rp/relying_party.go#L454) when a token response is received from the OP. If we were to accept this option on the Verifyr, then the RP should have it also.

You say GCP doesn't work with the azp claim? With that do you mean `accounts.google.com`, as that is the front for GCP login, right? We've been offering Google SSO inside zitadel for a long while now, with the default Relying Party (and azp check) and never had this problem.

Are you there is no underlying issue you are working around with this "fix"?

Joerger:
> This is a slight breaking change. Making the AZP check optional and disabled by default. This allows a security risk being introduced into system which rely on the claim being checked already.
> 
> Also note that `VerifyTokens` is also [called by the `RelyinParty`](https://github.com/joerger/oidc/blob/c5c849a93c9fa0a07c6e03202b94b54baa5e804d/pkg/client/rp/relying_party.go#L454) when a token response is received from the OP. If we were to accept this option on the Verifyr, then the RP should have it also.

Default behavior is restored in the latest version. I also moved the `AZPVerifier` option into `IDTokenVerifier` where it can also be used in the relying party methods. 339457111ea6aaca65162c4c1c454cd5b64d6d27

> You say GCP doesn't work with the azp claim? With that do you mean `accounts.google.com`, as that is the front for GCP login, right? We've been offering Google SSO inside zitadel for a long while now, with the default Relying Party (and azp check) and never had this problem.
> 
> Are you there is no underlying issue you are working around with this "fix"?

We aren't using accounts.google.com authn for humans, we're using OIDC ID Tokens issued by the [metadata service](https://cloud.google.com/compute/docs/instances/verifying-instance-identity#payload). In this case, GCP sets the `azp` to the value of `sub` rather than the `aud` , which breaks the verification that enforces that `azp` is equal to the ClientID.

It may be that GCP is making an odd decision here with the IDTokens, but allowing clients to specify the `azp` validation logic is in line with the latest OIDC documentation anyways, given this [errata](https://bitbucket.org/openid/connect/src/b84078b1aeb694a79823f3de5a22315df700b22f/openid-connect-core-1_0.xml#lines-1933:1944):
```
This validation MAY include that when an azp (authorized party) Claim is present, the Client SHOULD verify that its client_id is the Claim Value.
```

pkg/oidc/verifier.go
```diff
@@ -130,19 +131,34 @@ func CheckAudience(claims Claims, clientID string) error {
...
+}
+
 // CheckAuthorizedParty checks azp (authorized party) claim requirements.
 //
 // If the ID Token contains multiple audiences, the Client SHOULD verify that an azp Claim is present.
-// If an azp Claim is present, the Client SHOULD verify that its client_id is the Claim Value.
+// If an azp Claim is present, the Client MAY verify that its client_id is the Claim Value.
 // https://openid.net/specs/openid-connect-core-1_0.html#IDTokenValidation
-func CheckAuthorizedParty(claims Claims, clientID string) error {
+func CheckAuthorizedParty(claims Claims, azp AZPVerifier) error {
```
muhlemmer, line 153: It would be great if we can prevent this breaking change in the function signature as well.

I suggest to add a new function `CheckAZPVerifier(claims Claims, azp AZPVerifier) error` the body of the current proposal.

`CheckAuthorizedParty(claims Claims, clientID string) error` can wrap the new function using a `clientID`.

muhlemmer (changes requested):

pkg/oidc/verifier.go
Joerger: Sorry for the delay, I've addressed this and more in b5d7f778a65fbc0f2b4102728ab5b1d05fb9008f

pkg/client/rp/verifier.go
```diff
@@ -57,7 +57,11 @@ func VerifyIDToken[C oidc.Claims](ctx context.Context, token string, v *IDTokenV
 		return nilClaims, err
 	}
 
-	if err = oidc.CheckAuthorizedParty(claims, v.ClientID); err != nil {
+	if v.AZP != nil {
+		v.AZP = oidc.DefaultAZPVerifier(v.ClientID)
+	}
```
muhlemmer, line 62: Sorry I didn't notice before. Assigning an AZP at this stage might result in a race condition if the verifier is used by multiple go routines, like in a HTTP server.

It is better to assign the Default in the constructor function:

```go
// NewIDTokenVerifier returns a oidc.Verifier suitable for ID token verification.
func NewIDTokenVerifier(issuer, clientID string, keySet oidc.KeySet, options ...VerifierOption) *IDTokenVerifier {
	v := &IDTokenVerifier{
		Issuer:   issuer,
		ClientID: clientID,
		KeySet:   keySet,
		Offset:   time.Second,
		Nonce: func(_ context.Context) string {
			return ""
		},
		AZP: oidc.DefaultAZPVerifier(clientID), // <--
	}

	for _, opts := range options {
		opts(v)
	}

	return v
}
```

Then. if the option is set, the default is replaced by the function passed in the Option. While retaining the default if it wasn't, and not creating a race condition.

muhlemmer (approved):

muhlemmer (approved):

muhlemmer:
hey @Joerger thanks for the contribution.

PR #753, opened 2025-06-04
feat: Request aware cookie handling

markylaing:
Hi :wave: firstly, thank you for this great library!

LXD is a hypervisor for system containers and VMs. It can be clustered to give a private cloud like experience.

When LXD is clustered, we have an issue wherein the authorization code flow fails if the host that receives the callback (receiver of `/oidc/callback`) is different from the host that begins the flow (receiver of `/oidc/login`; see https://github.com/canonical/lxd/issues/13644). This is because the cookie handler in the relying party has different hash and encryption keys for each cluster member. Thus, when the callback is executed, the relying party cannot decode the state or PKCE cookies and the callback fails.

As discussed in the linked issue, this can be mitigated via sticky sessions on an external load balancer. However, we are increasingly seeing LXD deployments placed behind DNS SRV records, where sticky sessions are not possible. 

We would like to be able to derive cookie encryption keys from some shared secret, so that the hash and encryption keys are the same on all cluster members. To obfuscate the secret we would like to use a nonce value during derivation. Ideally this nonce would be unique per client.

The relying party accepts a `CookieHandler`, which in turn accepts static hash and block keys. In order to use unique keys per-client (or on rotation), a new relying party must be created for each cookie handler. This is not ideal, because creation of a new relying party re-triggers OIDC discovery and creates a new remote key set (which will retrieve JWKs when required). 

This PR adds functionality to the CookieHandler to allow more dynamic handling of cookies, based on the incoming request. I have taken care not to break any existing APIs.

Please let me know what you think and what my next steps should be (e.g. testing)

### Definition of Ready

- [ ] Acceptance criteria are met (not sure)
- [ ] All open todos and follow ups are defined in a new ticket and justified (not sure)
- [ ] Deviations from the acceptance criteria and design are agreed with the PO and documented. (not sure)
- [x] No debug or dead code
- [ ] My code has no repetitions (some slight repetition to avoid API breakage)
- [ ] Critical parts are tested automatically (not sure)
- [ ] All non-functional requirements are met (not sure)
- [ ] Functionality of the acceptance criteria is checked manually on the dev system. (not sure)

### DIFF
--- a/pkg/client/integration_test.go
+++ b/pkg/client/integration_test.go
@@ -3,6 +3,9 @@ package client_test
 import (
 	"bytes"
 	"context"
+	"crypto/hmac"
+	cryptoRand "crypto/rand"
+	"crypto/sha512"
 	"fmt"
 	"io"
 	"log/slog"
@@ -18,6 +21,8 @@ import (
 	"testing"
 	"time"
 
+	"github.com/google/uuid"
+	"github.com/gorilla/securecookie"
 	"github.com/jeremija/gosubmit"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
@@ -42,6 +47,15 @@ var Logger = slog.New(
 
 var CTX context.Context
 
+type cookieSpec struct {
+	cookieHandler *httphelper.CookieHandler
+	extraCookies  []*http.Cookie
+}
+
+var defaultCookieSpec = cookieSpec{
+	cookieHandler: httphelper.NewCookieHandler([]byte("test1234test1234"), []byte("test1234test1234"), httphelper.WithUnsecure()),
+}
+
 func TestMain(m *testing.M) {
 	os.Exit(func() int {
 		ctx, cancel := signal.NotifyContext(context.Background(), os.Interrupt, syscall.SIGINT)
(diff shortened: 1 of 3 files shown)


markylaing:
> Code looks great. Perhaps you can get some testing in by using (or creating a variant of) [testRelyingPartySession](https://github.com/zitadel/oidc/blob/7d57aaa99983368527b0fc93f72ab5542b2eefd2/pkg/client/integration_test.go#L63)

Great stuff I'll look into doing that and ping you when I have something :)

markylaing:
@muhlemmer hi :wave: I've added integration tests now. Tests all ok locally :)

pkg/client/integration_test.go
```diff
@@ -3,6 +3,9 @@ package client_test
 import (
 	"bytes"
 	"context"
+	"crypto/hkdf"
```
muhlemmer, line 6: Package is only available from go 1.24 onward. We support the latest 2 versions of go, that means 1.23.

muhlemmer (changes requested):

pkg/client/integration_test.go
markylaing: Apologies for the delay - I've updated the PR to use HMAC directly rather than HKDF :)

markylaing:
> Thanks @markylaing, Tim is on hols atm, he's back in a couple of days, then he'll have a look again!

Great stuff thank you :pray:

muhlemmer (approved):

PR #768, opened 2025-07-11
Issue/10239

migregal:
Follows https://github.com/zitadel/zitadel/issues/10239

### Definition of Ready

- [ ] No debug or dead code


pkg/op/server_legacy.go
```diff
@@ -177,6 +186,20 @@ func (s *LegacyServer) DeviceAuthorization(ctx context.Context, r *ClientRequest
...
+	defer span.End()
+
+	res, err := createPushedAuthorizationRequest(ctx, r.Data, s.provider)
+	if err != nil {
+		return nil, err
+	}
+
+	return NewResponseWithStatus(http.StatusCreated, res), nil
+}
+
```
muhlemmer, line 202: Does not need to be handled by `LegacyServer`, because it is a new feature, not one that needs to be kept as backward compatible.

migregal: @muhlemmer, can u share your thoughts on this, please?

migregal: Just tried to implement feature this way and it seems like a big enough refactoring issue.

Due to direct usage of Legacy server in Auth flow, there is problem with separating logic - LegacyServer has to know about PAR storage and feature itself, to correctly impl Auth handler. Alternative way is to refactor Auth logic and move it to webServer as well, but this will break backward compatibility.

Waiting for your answer here, but I think, the best way to impl PAR is to reuse LegacyServer for know

PR #791, opened 2025-08-31
fix: consistently handle string-valued boolean fields from non-compliant OIDC providers

bashhack:
### Background

👋 @muhlemmer - hope all is well with you and the team!

AWS Cognito (and potentially other providers) return `email_verified` and `phone_number_verified` as strings (`"true"`/`"false"`) instead of proper JSON booleans, violating the [OIDC specification](https://openid.net/specs/openid-connect-basic-1_0.html#StandardClaims).

AWS Documentation confirms this:
> Currently, Amazon Cognito returns the values for email_verified and phone_number_verified as strings.

_Source: https://docs.aws.amazon.com/cognito/latest/developerguide/userinfo-endpoint.html#get-userinfo-response-sample_

### The Problem

The `zitadel/oidc` library currently handles this inconsistently:
  - ✅ `EmailVerified` uses the custom `Bool` type (added in #139)
  - ❌ `PhoneNumberVerified` uses Go's standard `bool`
  
This forces developers to handle semantically identical fields differently:

```go
// Currently inconsistent code path
userInfo.EmailVerified = oidc.Bool(emailValue)    // Cast
userInfo.PhoneNumberVerified = phoneValue      // No cast
```

Additionally, the existing `Bool.UnmarshalJSON` implementation meant that false values couldn't overwrite true.

### Solution

Applied `Bool` type consistently to both fields and simplified `Bool.UnmarshalJSON` using a direct switch statement to:

  - Handle standard JSON booleans (true/false)
  - Handle AWS Cognito string format ("true"/"false")
  - Return errors on invalid input instead of silently failing
  - Allow false to overwrite true

 Updated tests to match codebase conventions, as well.

 ### Impact

`PhoneNumberVerified` changes from `bool` to `Bool` (type alias of `bool`). Most consumer code should work as-is since `Bool` is just a type alias. Direct type assertions would need updating.

### Definition of Ready

- [X] I am happy with the code
- [X] Short description of the feature/issue is added in the pr description
- [X] Acceptance criteria are met
- [X] No debug or dead code
- [X] My code has no repetitions
- [X] Critical parts are tested automatically
- [x] Where possible E2E tests are implemented
- [X] Documentation/examples are up-to-date
- [x] Functionality of the acceptance criteria is checked manually on the dev system.


pkg/oidc/userinfo.go
```diff
@@ -1,5 +1,50 @@
...
+			return nil
+		case "false":
+			*bs = false
+			return nil
+		}
+	}
+
+	return fmt.Errorf("cannot unmarshal %s into Bool", data)
+}
+
```
muhlemmer, line 47: I don't understand why we need a more verbose version of the Unmarshal method? The previous method did what it's supposed to do, right?

1. Read the bytes.
2. Check if it's quoted (`"true"`) or unquoted (`true`). Both cases set the `*b` to `true`.
3. Anything not `true` is false, nothing need to be set.

muhlemmer:
> Breaking Changes

Please don't use this string in your PR comment. It get's converted to the commit message later and it will trigger the semrel tool to create a new major release.

pkg/oidc/userinfo.go
bashhack: Thank you so much for the quick review!

Removed the `Breaking Changes` language from the PR comment + semver tooling interaction - will keep that in mind 🤝 

Can definitely give a bit of insight into my thinking around the UnmarshalJSON, in general, as well.

The initial insight I was hoping to address was that the original implementation didn't closely adhere to the `json.Unmarshaler` interface [contract](https://pkg.go.dev/encoding/json#Unmarshaler), in which we might expect to return an error if the JSON is invalid for the target type.

```go
// Original - always returns nil, even for invalid JSON
func (bs *Bool) UnmarshalJSON(data []byte) error {
      if string(data) == "true" || string(data) == `"true"` {
          *bs = true
      }
      return nil  // Contract violation - should error on invalid input
}
```

As I was testing, it appeared that one could end up silently accepting invalid JSON without errors to the caller:

```go
var b Bool
err := json.Unmarshal([]byte(`123`), &b)        // err = nil (should error)
err = json.Unmarshal([]byte(`"maybe"`), &b)     // err = nil (should error)
err = json.Unmarshal([]byte(`[1,2,3]`), &b)     // err = nil (should error)
```

My concern had been that this could mask unintended errors or make debugging extremely difficult - malformed data gets silently ignored rather than caught.

As a side effect of this investigation, I stumbled on the realization that the original method seems to have the potential for a more serious issue: values can't overwrite true. Once a Bool field is true, unmarshaling false or "false" wouldn't change it.

Here's what I noticed:

```go
// Original implementation
func (bs *Bool) UnmarshalJSON(data []byte) error {
      if string(data) == "true" || string(data) == `"true"` {
          *bs = true
      }
      return nil  // Never modifies *bs for false values!
}
```

Such that a caller could get into this state:

```go
user := UserInfo{EmailVerified: true}

// Provider sends update revoking verification
json.Unmarshal([]byte(`{"email_verified": false}`), &user)

// BUG: user.EmailVerified is still true!
// The false value was silently ignored
```

In practice - this may not be a scenario we're terribly concerned about, but my thinking was that if a caller did intend to (in some security-critical flow) revoke verification for either email or phone, my change could help ensure predictable behavior.

With all this said, would you prefer a simpler approach that still maintains the interface contract? 

Happy to iterate, as we could do something like this and still I think this could be a good balance, addressing my concerns but balancing with the spirit of the original code's concise form.

```go
// Proposal for your consideration...
func (bs *Bool) UnmarshalJSON(data []byte) error {
      s := string(data)
      switch s {
      case "true", `"true"`:
          *bs = true
      case "false", `"false"`:
          *bs = false
      default:
          return fmt.Errorf("cannot unmarshal %s into Bool", s)
      }
      return nil
}
```

bashhack:
@muhlemmer + @elinashoko - should be all set here, I think the end approach is a nice balance, thanks again! PR feedback implemented, and branch updated accordingly with `main` 🤝 

Congrats on the 1.25+ support, too! 🎉

PR #803, opened 2025-09-30
fix: add auth_methods client_secret_basic for refresh token and revok…

kufd:
I have auth provider which supports only auth_methods client_secret_basic

I found that in this case refresh token and revoke token requests do not works
Because basic auth header missed in appropriate http request

Here i try to fit is

I am happy to here feedback and adjust the PT


kufd:
@muhlemmer Could you please take a look at the PR description?
I’ve updated it — I hope it looks better now.

PR #809, opened 2025-10-20
fix(tracing): renamed tracer in op

### LINKED ISSUE #808: Add ability to disable or override tracer in pkg/op (private variable makes it impossible)
### Preflight Checklist

- [x] I could not find a solution in the existing issues, docs, nor discussions

### Describe your problem

# Problem
Hello, we want to use this library in a large financial institution. Tracing is enabled too aggressively inside github.com/zitadel/oidc/v3/pkg/op, but unlike the pkg/client package, the tracer there is a private variable.
Because of this, we cannot override or disable tracing without using unsafe hacks (go:linkname), which is not acceptable for production in regulated environments.
Example:
- In github.com/zitadel/oidc/v3/pkg/client, the tracer is public and can be replaced.
- In github.com/zitadel/oidc/v3/pkg/op, the tracer is private, so there is no official way to disable or customize it.

# Why this is a problem
Many companies (banks, enterprise, government) have strict policies about:
Disabling unnecessary tracing
Controlling telemetry centrally
Avoiding unsafe code and fork maintenance
Right now, the only way to disable tracing inside pkg/op is to patch it manually or use unsafe.

# Suggested solution
Any of the following would solve the issue cleanly:
## Option 1 (preferred)
Make the tracer exported:
```
var Tracer = ...
```
## Option 2
Expose a function:
```
func SetTracer(trace.Tracer)
```
## Option 3
Expose a function to configure or disable tracing:
```
func DisableTracing() 
// or
func ConfigureTracing(provider trace.TracerProvider)
```
# Why this change is useful
Keeps API backward-compatible
Removes the need for unsafe or custom forks
Makes behavior consistent with pkg/client
Allows regulated environments to integrate safely

### Version

v3.45.0

### Environment

Self-hosted

### Additional Context

_No response_

valerijhegaj:
Closes #808 

- Made the Tracer public to allow users to customize it, similar to how it’s done in the client.

- Private global var tracer renamed to public global var Tracer


muhlemmer:
Pleas use a proper PR title: https://github.com/zitadel/oidc/blob/main/CONTRIBUTING.md#make-a-pr

And a description that explains what is being fixed and how.

PR #873, opened 2026-04-16
fix: URL-encode client credentials in Basic Auth per RFC 6749 §2.3.1

sza2587:
## Summary

Four `Auth()` methods call `req.SetBasicAuth()` with raw client credentials. Per [RFC 6749 Section 2.3.1](https://datatracker.ietf.org/doc/html/rfc6749#section-2.3.1), client credentials MUST be encoded using the `application/x-www-form-urlencoded` encoding algorithm before being sent via HTTP Basic Authentication.

`net/http.SetBasicAuth` only base64-encodes the credentials — it does not URL-encode them first. When a client secret contains characters like `%`, authorization servers that URL-decode per the RFC (e.g., Keycloak) fail with errors like `URLDecoder: Incomplete trailing escape (%) pattern`.

The existing `AuthorizeBasic()` helper in `pkg/http/http.go` already correctly applies `url.QueryEscape`. This PR applies the same encoding to the four `Auth()` methods that were missing it:

- `RefreshTokenRequest.Auth` (`pkg/client/rp/relying_party.go`)
- `RevokeRequest.Auth` (`pkg/client/client.go`)
- `DeviceAccessTokenRequest.Auth` (`pkg/client/client.go`)
- `ClientCredentialsRequest.Auth` (`pkg/oidc/token_request.go`)

## Related

This is distinct from #857 / #858 (duplicate credentials issue), though both stem from the same `Auth()` methods. This PR fixes the encoding bug without removing the `ClientSecretBasicAuthRequest` interface, so it is not a breaking change.

## Testing

All existing tests pass. The fix is a one-line change per method, consistent with the pattern already used by `AuthorizeBasic()`.

### DIFF
--- a/pkg/client/client.go
+++ b/pkg/client/client.go
@@ -156,7 +156,7 @@ type RevokeRequest struct {
 
 func (r RevokeRequest) Auth(req *http.Request) {
 	if r.ClientSecret != "" {
-		req.SetBasicAuth(r.ClientID, r.ClientSecret)
+		req.SetBasicAuth(url.QueryEscape(r.ClientID), url.QueryEscape(r.ClientSecret))
 	}
 }
 
@@ -274,7 +274,7 @@ type DeviceAccessTokenRequest struct {
 
 func (r *DeviceAccessTokenRequest) Auth(req *http.Request) {
 	if r.ClientSecret != "" {
-		req.SetBasicAuth(r.ClientID, r.ClientSecret)
+		req.SetBasicAuth(url.QueryEscape(r.ClientID), url.QueryEscape(r.ClientSecret))
 	}
 }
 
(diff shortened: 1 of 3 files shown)


sza2587:
@muhlemmer Would appreciate a review when you get a chance. This is a small fix (one-line change per method) that's causing production issues for us — Keycloak crashes with `URLDecoder: Incomplete trailing escape (%) pattern` when client secrets contain `%`, because the `Auth()` methods don't URL-encode before `SetBasicAuth`. The fix matches the existing pattern in `AuthorizeBasic()`. Thanks!

wim07101993 (approved):
Hi, thank you for the contribution. Looks good, will merge.

PR #874, opened 2026-04-17
fix: remove override error after GetRefreshTokenInfo

suqin-haha:
# Which Problems Are Solved

all the error from `GetRefreshTokenInfo` falls to server error. 
because the error got wrapped the on the server error
```
RevocationRequestError(w, r, oidc.ErrServerError().WithParent(err))
```

there is `DefaultToServerError` under `RevocationRequestError`, that it defaults to oidc server err if it's not oidc error. We don't need to override the error to server error from `GetRefreshTokenInfo`
```
func RevocationError(err error) StatusError {
	e := oidc.DefaultToServerError(err, err.Error())
```


suqin-haha:
unnessasery.

PR #875, opened 2026-04-18
fix(op): allow dynamic loopback ports in post_logout_redirect_uri per RFC 8252 §7.3

### LINKED ISSUE #6615: allow dynamic port binding for native apps in end session
## User story

Allow dynamic port binding for native apps that use loop back addresses. The behavior is outlined [RFC8252, section 7.3](https://datatracker.ietf.org/doc/html/rfc8252#section-7.3)  but normally only applies to Authorization requests within the Oauth2 universe.

End-session is an OIDC-only endpoint specified in [OpenID Connect RP-Initiated Logout 1.0](https://openid.net/specs/openid-connect-rpinitiated-1_0.html) and therefore we did not take RFC8252 into account for the `post_logout_redirect_uri` parameter.

### Acceptance criteria

- [ ] Implement [RFC8252, section 7.3](https://datatracker.ietf.org/doc/html/rfc8252#section-7.3) for the `post_logout_redirect_uri` parameter in the End-Session endpoint
- [ ] Submit a suggestion to the OpenID foundation to clarify the RP Initiated Logout specifications wrt to Native apps.

----

### Preflight Checklist

- [X] I could not find a solution in the documentation, the existing issues or discussions
- [X] I have joined the [ZITADEL chat](https://zitadel.com/chat)

### Environment

ZITADEL Cloud

### Version

_No response_

we get post_logout_redirect_uri invalid when calling end_session endpoint for native apps.

### To reproduce

when adding a native (desktop) app to zitadel, it asks for the redirect urls and the post logout redirect url.
we set them both to the same url: `http://127.0.0.1`.
this works fine for authentication, but when we logout using the `end_session_endpoint`, we get 
```json
{
  "error": "invalid_request",
  "error_description": "post_logout_redirect_uri invalid"
}
```
when we pass the uri as: `http://127.0.0.1:12345` where `12345` is a random available port.
### Screenshots

_No response_

### Expected behavior

_No response_

### Operating System

_No response_

### Relevant Configuration

_No response_

### Additional Context

_No response_

ahmednfwela:
# Which Problems Are Solved

- The end-session endpoint rejects valid `post_logout_redirect_uri` values from native (desktop) apps that use the loopback interface with a dynamic port. Registering `http://127.0.0.1/callback` and logging out via `http://127.0.0.1:54321/callback` returns `{"error":"invalid_request","error_description":"post_logout_redirect_uri invalid"}`, even though the same client can authenticate via that exact loopback URI without issue.
- This breaks RP-initiated logout for any native app that follows [RFC 8252 §7.3](https://datatracker.ietf.org/doc/html/rfc8252#section-7.3) ("Loopback Interface Redirection"), which states that the authorization server *MUST* allow any port to be specified at request time for loopback IP redirect URIs.
- The asymmetry exists because the auth-request side already implements RFC 8252 §7.3 for `redirect_uri` (see `validateAuthReqRedirectURINative`), but `ValidateEndSessionPostLogoutRedirectURI` was a plain string-equality check.

# How the Problems Are Solved

- `ValidateEndSessionPostLogoutRedirectURI` in `pkg/op/session.go` gains a native-app loopback branch placed between the existing exact-match loop and the existing `HasRedirectGlobs` (DevMode) branch.
- The new branch reuses the helpers already used by the auth-request validator: `HTTPLoopbackOrLocalhost` (parses an http/https URI and reports whether the host is `127.0.0.1`, `[::1]`, or `localhost`) and `equalURI` (compares only path + raw query). When the client is `ApplicationTypeNative` and the incoming `post_logout_redirect_uri` resolves to a loopback host, each registered URI is parsed the same way and compared without considering the port — matching the auth-side semantics exactly.
- No change to public API. No change to the Client/HasRedirectGlobs interfaces.

# Additional Changes

- Adds `pkg/op/session_test.go` with table-driven coverage for `ValidateEndSessionPostLogoutRedirectURI`. The table covers: existing exact / no-match / glob cases (regression guard), native loopback v4/v6/localhost dynamic ports, native HTTPS loopback, native cross-loopback equivalence (v4 ↔ v6, inherited from `equalURI`), wrong-path negatives, web/user-agent dynamic-port rejection, and the literal RFC 8252 §7.3 example URIs (`http://127.0.0.1:51004/oauth2redirect/example-provider`, `http://[::1]:61023/oauth2redirect/example-provider`) plus the §8.3 `localhost` variant. Total: 24 cases, all passing.

# Additional Context

- Closes zitadel/zitadel#6615 (filed 2023-09-25).
- Mirrors the existing native-app handling in `pkg/op/auth_request.go` (`validateAuthReqRedirectURINative`, lines 352-382 in v3.47.0) so that auth and logout treat loopback URIs consistently.

### DIFF
--- a/pkg/op/session.go
+++ b/pkg/op/session.go
@@ -117,6 +117,19 @@ func ValidateEndSessionPostLogoutRedirectURI(postLogoutRedirectURI string, clien
 			return nil
 		}
 	}
+	// For native clients, RFC 8252 section 7.3 allows the use of dynamic ports
+	// on the loopback interface (127.0.0.1 / ::1 / localhost). Match the incoming
+	// post_logout_redirect_uri against registered loopback URIs while ignoring the port.
+	if client.ApplicationType() == ApplicationTypeNative {
+		if parsedURL, isLoopback := HTTPLoopbackOrLocalhost(postLogoutRedirectURI); isLoopback {
+			for _, uri := range client.PostLogoutRedirectURIs() {
+				registeredURL, ok := HTTPLoopbackOrLocalhost(uri)
+				if ok && equalURI(parsedURL, registeredURL) {
+					return nil
+				}
+			}
+		}
+	}
 	if globClient, ok := client.(HasRedirectGlobs); ok {
 		for _, uriGlob := range globClient.PostLogoutRedirectURIGlobs() {
 			isMatch, err := path.Match(uriGlob, postLogoutRedirectURI)
--- a/pkg/op/session_test.go
+++ b/pkg/op/session_test.go
@@ -1,0 +1,247 @@
+package op_test
+
+import (
+	"testing"
+
+	"github.com/golang/mock/gomock"
+	"github.com/stretchr/testify/assert"
+
+	"github.com/zitadel/oidc/v3/pkg/op"
+	"github.com/zitadel/oidc/v3/pkg/op/mock"
+)
+
+func TestValidateEndSessionPostLogoutRedirectURI(t *testing.T) {
+	tests := []struct {
+		name                  string
+		registered            []string
+		registeredGlobs       []string
+		applicationType       op.ApplicationType
+		postLogoutRedirectURI string
+		useGlobsClient        bool
+		wantErr               bool
+	}{
+		{
+			name:                  "exact match (web)",
+			registered:            []string{"https://logged-out"},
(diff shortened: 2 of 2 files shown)


wim07101993 (approved):

wim07101993:
Thanks for the contribution. Looks good, will merge.

PR #902, opened 2026-06-18
fix(op): add client ownership check to LegacyServer CodeExchange

canolgun:
LegacyServer.CodeExchange looks up the auth request by code only, without checking that the authenticated client owns it.

The modern path in ValidateAccessTokenRequest (token_code.go:59) checks:
  if client.GetID() != authReq.GetClientID() { return error }

This adds the same check to LegacyServer.CodeExchange.


wim07101993 (commented):
Hi, thank you for the contribution. Could you update the pr title to match a semantic PR title

wim07101993 (changes requested):

wim07101993 (approved):

PR #921, opened 2026-07-16
refactor: use global slog logger

### LINKED ISSUE #829: Remove logging library references
Currently the OIDC server uses it own `slog` logger instance, based on our [logging library](https://github.com/zitadel/logging). After https://github.com/zitadel/zitadel/pull/11159, the library will be deprecated. OIDC should use the global slog loggers instead. Zitadel already populates context data in middleware, so this does not have to be handled by OIDC anymore.

## Acceptance criteria

- [ ] Use the `slog` global logging functions
- [ ] Remove all references to https://github.com/zitadel/logging
- [ ] Update documentation / readme and explain the global / default logger is now used and how to disable that if not wished.

sukujgrg:
# Which Problems Are Solved

- The OIDC server maintains provider-owned logger instances instead of using the global `slog` logger configured by applications.
- The project depends on `github.com/zitadel/logging`, which is being deprecated.
- Context-aware logging is duplicated even though applications can populate context data through their middleware.

# How the Problems Are Solved

- Replaces provider-owned logging calls with global `slog` functions.
- Uses the default logger configured by the application.
- Removes all references to and the dependency on `github.com/zitadel/logging`.
- Retains the existing logger options as deprecated no-ops for source compatibility.
- Documents how to configure or disable logging through `slog.SetDefault`.

# Additional Changes

- Removes the remaining `logrus` references and dependency.
- Updates examples to use `slog`.
- Adds upgrading guidance for applications using the deprecated logger options.

# Additional Context

- Closes #829


pkg/client/rp/relying_party.go
```diff
@@ -191,12 +189,10 @@ func (rp *relyingParty) UnauthorizedHandler() func(http.ResponseWriter, *http.Re
...
-func (rp *relyingParty) Logger(ctx context.Context) (logger *slog.Logger, ok bool) {
-	logger, ok = logging.FromContext(ctx)
-	if ok {
-		return logger, ok
-	}
-	return rp.logger, rp.logger != nil
+// Logger is retained for source compatibility.
+// Deprecated: configure logging with [slog.SetDefault]. It always returns nil, false.
+func (rp *relyingParty) Logger(context.Context) (*slog.Logger, bool) {
+	return nil, false
```
wim07101993: This will break current implementations. Please keep the old behavior. We can remove the functionality later.

I rather like the implementation you did on `op.Provider`, there you return `slog.Default()`

pkg/client/rp/relying_party.go
```diff
@@ -393,11 +388,10 @@ func WithJWTProfile(signerFromKey SignerFromKey) Option {
...
 
-// WithLogger sets a logger that is used
-// in case the request context does not contain a logger.
-func WithLogger(logger *slog.Logger) Option {
-	return func(rp *relyingParty) error {
-		rp.logger = logger
+// WithLogger is retained for source compatibility.
+// Deprecated: use [slog.SetDefault]. This option has no effect.
+func WithLogger(*slog.Logger) Option {
+	return func(*relyingParty) error {
```
wim07101993: This will break current implementations. Please keep the old behavior. We can remove the functionality later.

pkg/op/error.go
```diff
@@ -28,20 +28,20 @@ type LogAuthRequest interface {
 
 func AuthRequestError(w http.ResponseWriter, r *http.Request, authReq ErrAuthRequest, err error, authorizer Authorizer) {
 	e := oidc.DefaultToServerError(err, err.Error())
-	logger := authorizer.Logger().With("oidc_error", e)
+	args := []any{"oidc_error", e.LogValue()}
```
wim07101993: i think it would be clearer to use [slog attributes](https://pkg.go.dev/log/slog#Attr). That way it is clear which args are keys and which are values.

Also by calling `e.LogValue()` this value will always be caclulated, even when the log is omitted due to its level. If I am not mistaking `e.LogValue()` is called automatically by slog because it implements `LogValuer`

pkg/op/error.go
```diff
@@ -75,28 +76,28 @@ func RequestError(w http.ResponseWriter, r *http.Request, err error, logger *slo
...
 }
 
 // TryErrorRedirect tries to handle an error by redirecting a client.
 // If this attempt fails, an error is returned that must be returned
 // to the client instead.
-func TryErrorRedirect(ctx context.Context, authReq ErrAuthRequest, parent error, encoder httphelper.Encoder, logger *slog.Logger) (*Redirect, error) {
+func TryErrorRedirect(ctx context.Context, authReq ErrAuthRequest, parent error, encoder httphelper.Encoder, _ *slog.Logger) (*Redirect, error) {
 	e := oidc.DefaultToServerError(parent, parent.Error())
-	logger = logger.With("oidc_error", e)
+	args := []any{"oidc_error", e.LogValue()}
```
wim07101993: i think it would be clearer to use [slog attributes](https://pkg.go.dev/log/slog#Attr). That way it is clear which args are keys and which are values.

pkg/op/error_test.go
```diff
@@ -17,6 +17,13 @@ import (
 	"github.com/zitadel/schema"
 )
 
+func setDefaultTestLogger(t *testing.T, logger *slog.Logger) {
```
wim07101993, line 22: This function will break tests which run in parallel. Maybe add a comment so that this is clear.

wim07101993 (changes requested):
Thank you for the contribution. Nice work, I have some small comments though.

pkg/op/error.go
```diff
@@ -28,20 +28,20 @@ type LogAuthRequest interface {
 
 func AuthRequestError(w http.ResponseWriter, r *http.Request, authReq ErrAuthRequest, err error, authorizer Authorizer) {
 	e := oidc.DefaultToServerError(err, err.Error())
-	logger := authorizer.Logger().With("oidc_error", e)
+	args := []any{"oidc_error", e.LogValue()}
```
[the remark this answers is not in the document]
sukujgrg: fixed in 706ff6d. Replaced the eager `e.LogValue()` call with `slog.Any("oidc_error", e)`, so `slog` resolves the `LogValuer` at log time after State and SessionState have been populated. I also updated the log-output tests to verify the state field.

pkg/op/error.go
```diff
@@ -75,28 +76,28 @@ func RequestError(w http.ResponseWriter, r *http.Request, err error, logger *slo
...
 }
 
 // TryErrorRedirect tries to handle an error by redirecting a client.
 // If this attempt fails, an error is returned that must be returned
 // to the client instead.
-func TryErrorRedirect(ctx context.Context, authReq ErrAuthRequest, parent error, encoder httphelper.Encoder, logger *slog.Logger) (*Redirect, error) {
+func TryErrorRedirect(ctx context.Context, authReq ErrAuthRequest, parent error, encoder httphelper.Encoder, _ *slog.Logger) (*Redirect, error) {
 	e := oidc.DefaultToServerError(parent, parent.Error())
-	logger = logger.With("oidc_error", e)
+	args := []any{"oidc_error", e.LogValue()}
```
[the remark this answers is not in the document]
sukujgrg: same as above `AuthRequestError` fix.

pkg/op/token_code.go
```diff
@@ -17,20 +17,20 @@ func CodeExchange(w http.ResponseWriter, r *http.Request, exchanger Exchanger) {
 
 	tokenReq, err := ParseAccessTokenRequest(r, exchanger.Decoder())
 	if err != nil {
-		RequestError(w, r, err, exchanger.Logger())
+		RequestError(w, r, err, nil)
 	}
```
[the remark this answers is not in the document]
sukujgrg: Fixed in 706ff6d. Added an immediate return after writing the parse-error response, preventing tokenReq from being dereferenced when parsing fails. This issue already existed on main, but I addressed it here because this PR touched the error-handling path.

pkg/op/token_refresh.go
```diff
@@ -30,16 +30,16 @@ func RefreshTokenExchange(w http.ResponseWriter, r *http.Request, exchanger Exch
 
 	tokenReq, err := ParseRefreshTokenRequest(r, exchanger.Decoder())
 	if err != nil {
-		RequestError(w, r, err, exchanger.Logger())
+		RequestError(w, r, err, nil)
 	}
```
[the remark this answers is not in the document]
sukujgrg: same as above `ParseAccessTokenRequest` fix.

pkg/op/token_client_credentials.go
```diff
@@ -18,18 +18,18 @@ func ClientCredentialsExchange(w http.ResponseWriter, r *http.Request, exchanger
 
 	request, err := ParseClientCredentialsRequest(r, exchanger.Decoder())
 	if err != nil {
-		RequestError(w, r, err, exchanger.Logger())
+		RequestError(w, r, err, nil)
 	}
```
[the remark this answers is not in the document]
sukujgrg: same as above fix. returning immediately.

pkg/op/token_jwt_profile.go
```diff
@@ -22,23 +22,23 @@ func JWTProfile(w http.ResponseWriter, r *http.Request, exchanger JWTAuthorizati
 
 	profileRequest, err := ParseJWTProfileGrantRequest(r, exchanger.Decoder())
 	if err != nil {
-		RequestError(w, r, err, exchanger.Logger())
+		RequestError(w, r, err, nil)
 	}
```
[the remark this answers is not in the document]
sukujgrg: same as above fix. returning immediately.

pkg/op/token_exchange.go
```diff
@@ -140,17 +140,17 @@ func TokenExchange(w http.ResponseWriter, r *http.Request, exchanger Exchanger)
 
 	tokenExchangeReq, clientID, clientSecret, err := ParseTokenExchangeRequest(r, exchanger.Decoder())
 	if err != nil {
-		RequestError(w, r, err, exchanger.Logger())
+		RequestError(w, r, err, nil)
 	}
```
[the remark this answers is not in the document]
sukujgrg: same as above fix. returning immediately.

pkg/client/rp/relying_party.go
sukujgrg: Fixed in 706ff6d. Logger() now returns the logger configured through WithLogger, or slog.Default() when none is configured.

pkg/client/rp/relying_party.go
sukujgrg: Done in 706ff6d

pkg/op/error.go
sukujgrg: Fixed in 706ff6d. Replaced the key/value arguments with explicit slog.Attr values created using slog.Any, and passed the error directly so its LogValuer is evaluated automatically.

pkg/op/error.go
sukujgrg: same as above. replaced with explicit `slog.Attr` values created using `slog.Any`

pkg/op/error_test.go
sukujgrg: Done.

wim07101993 (approved):

PR #928, opened 2026-07-28
feat: support custom issuer verifier

suqin-haha:
# Which Problems Are Solved
Azure split issuer by tenent id `https://login.microsoftonline.com/{tenantid}/v2.0`
when we want support more general users, like both company and personal users, the issuer is dynamically different by the tenent. the discovery endpoint [https://login.microsoftonline.com/common/v2.0/.well-known/openid-configuration](https://login.microsoftonline.com/common/v2.0/.well-known/openid-configuration) shows the `{tenantid}` pattern as well. 

This PR support customized issuer verifier to support dynamic issuer. 

reference: 
https://learn.microsoft.com/en-us/entra/identity-platform/v2-protocols-oidc#find-your-apps-openid-configuration-document-uri

# How the Problems Are Solved

This PR support customized issuer verifier to support dynamic issuer. 

# Additional Changes

change all the usage to use the new function `oidc.CheckIssuerWithISSVerifier(claims, v.Issuer, v.ISS)` 
it does not change current or default usage, just add option to support cusomized issuer verifier. 

# Additional Context

### DIFF
--- a/pkg/client/rp/verifier.go
+++ b/pkg/client/rp/verifier.go
@@ -132,6 +132,7 @@ func NewIDTokenVerifier(issuer, clientID string, keySet oidc.KeySet, options ...
 // VerifierOption is the type for providing dynamic options to the IDTokenVerifier
 type VerifierOption func(*IDTokenVerifier)
 
+// WithISSVerifier sets a custom verifier for validating the `iss` (issuer) claim.
 func WithISSVerifier(verifier oidc.ISSVerifier) VerifierOption {
 	return func(v *IDTokenVerifier) {
 		v.ISS = verifier
--- a/pkg/client/rp/verifier.go
+++ b/pkg/client/rp/verifier.go
@@ -49,7 +49,7 @@ func VerifyIDToken[C oidc.Claims](ctx context.Context, token string, v *IDTokenV
 		return nilClaims, err
 	}
 
-	if err = oidc.CheckIssuer(claims, v.Issuer); err != nil {
+	if err := oidc.CheckIssuerWithISSVerifier(claims, v.Issuer, v.ISS); err != nil {
 		return nilClaims, err
 	}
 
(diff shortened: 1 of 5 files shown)


pkg/client/rp/verifier.go
```diff
@@ -49,7 +49,7 @@ func VerifyIDToken[C oidc.Claims](ctx context.Context, token string, v *IDTokenV
 		return nilClaims, err
 	}
 
-	if err = oidc.CheckIssuer(claims, v.Issuer); err != nil {
```
muir, line 52: Also deprecate CheckIssuer.

pkg/oidc/verifier.go
```diff
@@ -116,13 +131,21 @@ func CheckSubject(claims Claims) error {
...
 		return fmt.Errorf("%w: Expected: %s, got: %s", ErrIssuerInvalid, issuer, claims.GetIssuer())
 	}
 	return nil
 }
 
+func CheckIssuerWithISSVerifier(claims Claims, issuer string, ISS ISSVerifier) error {
+	if ISS == nil {
+		return DefaultISSVerifier(issuer)(claims.GetIssuer())
+	}
+	return ISS(claims.GetIssuer())
```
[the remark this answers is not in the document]
wim07101993: It's semantics but I agree, the ISS was a bit confusing in the beginning. So maybe rename it to `issVerifier`, just for clarity.

pkg/oidc/verifier.go
```diff
@@ -116,13 +131,21 @@ func CheckSubject(claims Claims) error {
 	return nil
 }
 
+// Deprecated: Use CheckIssuerWithISSVerifier instead
 func CheckIssuer(claims Claims, issuer string) error {
```
wim07101993, line 135: I don't actually know whether we should deprecate this. I think the implementation can call `CheckIssuerWithISSVerifier` with as verifier nil or the `DefaultISSVerifier`. Most people won't require a dynamic resolving of the issuer verifier, I think.

suqin-haha: um... I think maybe we should, since we already have `CheckIssuerWithISSVerifier` which cover all the cases. So the old function no longer needed anymore.  (we might could change the `CheckIssuerWithISSVerifier` to `CheckIssuer` or somthing more general name. )

with less code might cause less confuse and clean repo. 

feel free let me know if we want keep the Deprecated mark or remove it. I will do the change. 

Thanks for the quick review!

wim07101993 (approved):

PR #821, opened 2025-12-01
fix!: allow expires to be an integer wrapped in a string value

wim07101993:
Certain providers, for example Entra ID, return durations as quoted strings instead of JSON numbers. This is a deviation from the standard. This change allows for the successful parsing of such quoted duration strings.

BREAKING CHANGE: The `expires_in field` has changed from type `int64` to `oidc.Duration` for various token response payloads.

fixes https://github.com/zitadel/oidc/issues/815

### Definition of Ready

- [x] I am happy with the code
- [x] Short description of the feature/issue is added in the pr description
- [x] PR is linked to the corresponding user story
- [x] Acceptance criteria are met
- [x] All open todos and follow ups are defined in a new ticket and justified
- [ ] Deviations from the acceptance criteria and design are agreed with the PO and documented.
- [x] No debug or dead code
- [x] My code has no repetitions
- [x] Critical parts are tested automatically
- [ ] Where possible E2E tests are implemented
- [ ] Documentation/examples are up-to-date
- [ ] All non-functional requirements are met
- [ ] Functionality of the acceptance criteria is checked manually on the dev system.

PR #929, opened 2026-07-29
fix(oidc): callTokenEndpoint expiry conversion error

EthanHeilman:
# Which Problems Are Solved

A bug was introduced in the PR https://github.com/zitadel/oidc/pull/821 which changes `exp` claim in `AccessTokenResponse` from a `int64` to an `oidc.Duration` but does not update `callTokenEndpoint`. Thus, `callTokenEndpoint` still treats it as int64 and and multiples it by `time.second`. This results in an incorrect Expiry value, depending on the `expiresIn` value it can be ahead by 30 years all the way to dates well exceeding the time period of Futurama.

# How the Problems Are Solved

We fix this bug by updating `callTokenEndpoint` to bring in line the intent of #821 

From: `Expiry:       time.Now().UTC().Add(time.Duration(tokenRes.ExpiresIn) * time.Second),`
To: `Expiry:       time.Now().UTC().Add(tokenRes.ExpiresIn.AsDuration()),`

# Additional Changes

We also add a simple regression test for this bug.

# Additional Context

Note this that only impacts v4 and v3 since this change from `int64` to `oidc.Duration` only exists in v4. We encountered this bug when preparing OpenPubkey connect to update from v3 to v4 https://github.com/openpubkey/openpubkey/pull/392

### DIFF
--- a/pkg/client/client.go
+++ b/pkg/client/client.go
@@ -87,7 +87,7 @@ func callTokenEndpoint(ctx context.Context, request any, authFn any, caller Toke
 		AccessToken:  tokenRes.AccessToken,
 		TokenType:    tokenRes.TokenType,
 		RefreshToken: tokenRes.RefreshToken,
-		Expiry:       time.Now().UTC().Add(time.Duration(tokenRes.ExpiresIn) * time.Second),
+		Expiry:       time.Now().UTC().Add(tokenRes.ExpiresIn.AsDuration()),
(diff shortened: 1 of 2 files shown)


wim07101993:
Whoops, that was my mistake. Thanks for the fix. Will merge.

wim07101993 (approved):

PR #943, opened 2026-08-12
feat(oidc,client): Adds support for OpenID Key Binding to client

EthanHeilman:
# Which Problems Are Solved

This PR adds client support for [OpenID Key Binding 1.0](https://openid.net/specs/openid-connect-key-binding-1_0.html) a draft standard for Proof-of-Possession ID Tokens. Basically ID Tokens that have a public key enabling the user to sign under their identity.

OpenPubkey supports key binding using zitadel/oidc as the library. However because zitadel/oidc does not natively support key binding, OpenPubkey has to use [a fragile rounder-tripper work around for the RP](https://github.com/openpubkey/openpubkey/blob/main/providers/keybinding_provider.go#L109).

This PR allows key binding support for OP and RP:
- Authorization code flow
- Device code flow
- Refresh flow
- Unit tests

This PR is only for client/RP. It does not add key binding to the OP code. I would like to add that at a future point, but client/RP support the main blocker to using plain zitadel/oidc with key binding.

I've run manual tests for code flow, refresh and device flow using the key binding with the Hello OP and this PR in OpenPubkey that runs against this branch: https://github.com/openpubkey/openpubkey/pull/395

# How the Problems Are Solved

Adds a `WithKeyBinding` option to  RP so that the RP can request key bound ID Tokens from the OP and a `KeyBindingRelyingParty` to perform the additional steps used by key binding.

```golang
func WithKeyBinding(signer crypto.Signer, alg jose.SignatureAlgorithm) Option
```

# Additional Changes

No additional changes

# Additional Context

Do you want me to port this to v4 as well as v3?

Relevant PRs
- PR adding Key Binding to Authentik https://github.com/goauthentik/authentik/pull/22590
- [OpenID Key Binding 1.0 Latest Draft](https://openid.net/specs/openid-connect-key-binding-1_0.html)
- PR to use this RP option in OpenPubkey https://github.com/openpubkey/openpubkey/pull/395


EthanHeilman:
@wim07101993 How much interest is there in merging this? Do you want me to move this to v4 rather v3?

pkg/oidc/dpop.go
```diff
@@ -0,0 +1,122 @@
...
+
+// CanonicalJWK returns the public key members of jwk without optional or
+// caller-controlled JWK metadata such as kid, use, alg, or x5c.
+func CanonicalJWK(jwk *jose.JSONWebKey) (json.RawMessage, error) {
+	canonical, err := json.Marshal(jose.JSONWebKey{Key: jwk.Key})
+	if err != nil {
+		return nil, err
+	}
+	return json.RawMessage(canonical), nil
+}
```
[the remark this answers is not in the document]
EthanHeilman: This function was left over from earlier version of the PR and is not currently used. I've deleted it.

kipz:
@EthanHeilman it looks like we're not handling DPoP server provided nonces: https://www.rfc-editor.org/rfc/rfc9449.html#section-8 - not sure if this deliberate for now, but I think auth/resource servers sending them will fail if they're not received from the client.

EthanHeilman:
@kipz Thanks for the review!

We don't use a DPoP `nonce` as the [OpenID Key Binding draft standard](https://openid.net/specs/openid-connect-key-binding-1_0.html) does not use DPoP `nonce` for the ID Token issuance. Instead  the standard `code` as a DPoP `nonce`. This has two benefits:

- DPoP `nonce` is optional in OAuth2 token issuance, whereas since OpenID Key Binding uses `code` which exists in all supported key binding flows and so we can require it.
- It also binds the issued ID Token and Refresh Token to a specific authorization session.

The DPoP code in this PR is only for issuance.

kipz:
thanks @EthanHeilman , yeah, I see the key binding spec uses `c_s256` which covers session binding for the code and device flows.

one thing though: "optional" in [rfc 9449 §8](https://www.rfc-editor.org/rfc/rfc9449.html#section-8) reads to me as optional for the AS, not the client. If an AS does require a nonce it answers with `400 use_dpop_nonce` plus a `DPoP-Nonce` header, and [§11.3](https://www.rfc-editor.org/rfc/rfc9449.html#section-11.3) says it MUST then reject proofs without the claim. The spec also has the OP do all the [§5](https://www.rfc-editor.org/rfc/rfc9449.html#section-5) checks, so I don't think `c_s256` displaces that?

Guess it's more of a potential interop issue for later maybe :)

EthanHeilman:
RFC 9449 and OpenID Key Binding intentionally differ in this regard and they are different protocols. Key binding token issuance only requires [RFC 9449 Section 5](https://www.rfc-editor.org/info/rfc9449/#section-5) validation. When doing key binding, `c_s256` replaces the DPoP `nonce`. 

When using an key bound ID Token, nonce headers should respected by the client. When issuing a key bound ID Token, a nonce should not be required by the AS.

To rant for a second because it is a topic close to my heart =)

The use of a DPoP `nonce` as written in RFC 9449 introduces a lot of complexity. The protocol no longer has a set number of interactions but can continue indefinitely, i.e., AS says use nonce-1, client uses nonce-1 but AS rotates to nonce-2, rejects the client's DPoP proof because it doesn't use nonce-2, client tries nonce-2, but the AS has rotated to nonce-3, ... This is not an impossible outcome, consider the AS being load balanced between two availability zones based on latency, the AS may not share the same nonce across zones. Token issuance, which was before a simple problem, has now can become byzantine through the introduction of this nonce requirement.

From an implementers point of view, key binding is nice because it does not add any additional requests, it just augments the existing requests. If we added nonce's into the flow, we now have a "opps wrong nonce" response from the AS and a client "retrying with nonce" request.

Even with all that, DPoP nonces in this context don't even solve the problems we want to solve, replay protection and session binding. This is because the standard does not require that they are single use and unique to a session. We already have something in that is single use and unique to the session the authorization code.

In practice, as far as I can tell, no one uses nonces in access token issuance due to the above issues. Key binding provides a better pattern here.

kipz:
Thanks for the detailed response/rant @EthanHeilman! 🙇🏿‍♂️

kipz:
FWIW, it'd be super handy to get this merged so that https://github.com/openpubkey/openpubkey/pull/395 can go in! I'm waiting on this to move all my/our openpubkey use away from the aud/nonce claim to 'cnf' which would be amazing and a long time in the waiting!

muhlemmer:
Normally we're careful not to accept contributions of standards that are still in draft. We are willing to accept, but please mark all exported GO API as `Experimental`. So if the draft changes or is never published, we can update / remove without breaking semver.

pkg/client/client.go
```diff
@@ -248,6 +248,23 @@ type DeviceAuthorizationCaller interface {
...
+// BoundKeyDeviceAuthorizationRequest adds the `dpop_jkt` parameter from
+// OpenID Connect Key Binding 1.0, Section 3.1 to a Device Authorization Request.
+type BoundKeyDeviceAuthorizationRequest struct {
+	*oidc.ClientCredentialsRequest
+	DPoPJKT string `schema:"dpop_jkt,omitempty"`
+}
+
+// CallDeviceAuthorizationEndpointWithBoundKey is [CallDeviceAuthorizationEndpoint]
+// that includes a request with an OpenID Key Binding proof-of-possession key.
+func CallDeviceAuthorizationEndpointWithBoundKey(ctx context.Context, request *BoundKeyDeviceAuthorizationRequest, caller DeviceAuthorizationCaller, authFn any) (*oidc.DeviceAuthorizationResponse, error) {
```
wim07101993: Creating a separate function feels a bit smelly. Can't we do something with more like an option-shaped API?

wim07101993 (commented):
Hi, thank you for the contribution. I am sorry that it took this long to review. I had to get up to speed/refresh with keybinding and dpop. 

Overal looks good. Just one question.

pkg/client/client.go
EthanHeilman: Switched it to an option that specifies the JKT. Tested against OpenPubkey with device auth

EthanHeilman:
@muhlemmer Marked all exported go APIs as `Experimental`.

pkg/client/rp/key_binding.go
```diff
@@ -0,0 +1,336 @@
...
+	htu.ForceQuery = false
+	htu.Fragment = ""
+	htu.RawFragment = ""
+
+	// Only ever sign a proof for the configured token endpoint. Without this,
+	// a 307/308 redirect from the token endpoint would make net/http replay the
+	// POST body (authorization code and client secret) to the redirect target,
+	// and this transport would helpfully mint a fresh proof for that host,
+	// disclosing c_s256 = SHA256(code) to it.
+	if t.tokenEndpoint == "" || htu.String() != t.tokenEndpoint {
```
wim07101993: Shouldn't this comparison happen without query-string and fragments of the token-endpoint as well? 

If a token endpoint contains a query param like `tenant=my-tenant`, this will fail. Which happens with Microsoft Entry ID if I am not mistaking

wim07101993 (commented):
Just one more question. Sorry didn't see it the first time.

pkg/client/rp/key_binding.go
EthanHeilman: That's a really good catch, I could see someone putting that into their config and having a hard to diagnose bug. Fixed that and added some additional validation and error messages around it.

pkg/client/rp/key_binding.go
```diff
@@ -240,30 +249,53 @@ func (k *keyBinding) proof(method, tokenEndpoint, code string) (string, error) {
...
+}
+
+// htuKey returns a normalized version of the token endpoint URI
+// for comparison per RFC 3986, sections 6.2.2 and 6.2.3.
+func htuKey(u *url.URL) string {
+	n := *u
+	n.Scheme = strings.ToLower(n.Scheme)
+	n.Host = strings.ToLower(n.Host)
+	if port := n.Port(); (n.Scheme == "https" && port == "443") || (n.Scheme == "http" && port == "80") {
+		n.Host = n.Hostname()
```
wim07101993: AI comment, wouldn't have seen this myself:

IPv6 normalization bug: removing `:443`/`:80` via `n.Host = n.Hostname()` drops IPv6 brackets. `https://[::1]:443/token` becomes invalid `https://::1/token` and won’t match `https://[::1]/token`. Rebuild the host with brackets and add IPv6 default-port tests.

wim07101993 (commented):
I think with the normalization, a bug slipped in.

pkg/client/rp/key_binding.go
EthanHeilman: Fixed, I found the simplest way to do this is to simple trim suffix the port and left the rest of the host unchanged. Added a test to cover this case.

 I very much appreciate the level of code review in this project. It is impressive.

wim07101993 (approved):
Thank you for the contribution and patience. Will merge.

PR #955, opened 2026-08-23
feat(op): support resource indicators (RFC 8707) at the authorization endpoint

### LINKED ISSUE #794: Support of RFC-8707
### Preflight Checklist

- [x] I could not find a solution in the existing issues, docs, nor discussions

### Describe your problem

Current implementation of oidc.AuthRequest doesn't support `resource` as specified in [RFC-8707](https://datatracker.ietf.org/doc/html/rfc8707). This is useful for [MCP servers implementations](https://modelcontextprotocol.io/specification/draft/basic/authorization#resource-parameter-implementation).

Add support in [oidc.AuthRequest](https://github.com/zitadel/oidc/blob/044894f6860eff5b860678b8d749a96f813372ff/pkg/oidc/authorization.go#L69). Ideally the implementation would also add the resource parsing a middleware to validate inputs for `resource` values as the RFC states:

> resource
>
>    Indicates the target service or resource to which access is being requested. Its value MUST be an absolute URI, as specified by [Section 4.3](https://www.rfc-editor.org/rfc/rfc3986#section-4.3) of [[RFC3986](https://datatracker.ietf.org/doc/html/rfc3986)]. The URI MUST NOT include a fragment component. It SHOULD NOT include a query component, but it is recognized that there are cases that make a query component a useful and necessary part of the resource parameter, such as when one or more query parameters are used to scope requests to an application. The resource parameter URI value is an identifier representing the identity of the resource, which MAY be a locator that corresponds to a network-addressable location where the target resource is hosted. Multiple resource parameters MAY be used to indicate that the requested token is intended to be used at multiple resources.

Something like this would do the job:

```go
values := r.URL.Query()
for i := range values["resource"] {
	u, err := url.ParseRequestURI(values["resource"][i])
	if err != nil {
		// todo: render oauth error with `invalid_target`
	}

	if len(u.Fragment) > 0 {
		// todo: render oauth error with `invalid_target`
	}
}
```

### Version

_No response_

### Environment

Self-hosted

### Additional Context

Happy to make a PR if the implementation works for the team

amartya-dev:
# Which Problems Are Solved

- `oidc.AuthRequest` has no field for the `resource` parameter of [RFC 8707 (Resource Indicators for OAuth 2.0)](https://www.rfc-editor.org/rfc/rfc8707), so the parameter is silently dropped at the authorization endpoint. A `Storage` implementation never learns which resource a token was requested for, and therefore cannot bind the audience of the issued token to it. The same is true at the token endpoint and in the device authorization flow.
- The only workaround today is to read `resource` out of the raw `http.Request` before it reaches `op.Authorize` and thread it into the audience by hand. That is easy to get subtly wrong — the failure mode is a token with a wildcard or default audience, which defeats the point of asking for a resource in the first place.
- Nothing validates the syntax of `resource` values, and there is no way for an OP to advertise the capability in its discovery document.

The motivation is audience binding for MCP servers. [MCP revision 2026-07-28](https://modelcontextprotocol.io/specification/draft/basic/authorization#resource-parameter-implementation) makes every MCP server an OAuth 2.1 resource server: the server MUST validate that a token's audience is itself, and the client MUST send the RFC 8707 `resource` parameter. Resource indicators are the mechanism that audience binding depends on.

# How the Problems Are Solved

The PR is split into four commits, each of which builds and tests on its own, so they can be read (or split into separate PRs) one at a time.

### 1. `feat(op): support resource indicators (RFC 8707) at the authorization endpoint`

- `oidc.AuthRequest` gains `Resource []string` with the `resource` schema tag, so repeated `resource` parameters are decoded at `/authorize` and reach `Storage.CreateAuthRequest` on the `*oidc.AuthRequest` that is already passed to it. No interface change is required, so this is fully backwards compatible.
- `op.ValidateResourceIndicators` validates the values per [RFC 8707, section 2](https://www.rfc-editor.org/rfc/rfc8707#section-2): each value must be an absolute URI and must not include a fragment component. A query component is explicitly allowed, as the RFC permits. Invalid values are rejected with the `invalid_target` error code the RFC specifies.
- `CopyRequestObjectToAuthRequest` copies `Resource` from a Request Object, consistent with how every other authorization parameter is handled there.
- `oidc.DiscoveryConfiguration` gains `ResourceIndicatorsSupported` (`resource_indicators_supported`, `omitempty`), driven by a new `op.Config.ResourceIndicatorsSupported` field via the `op.ResourceIndicatorsSupported(Configuration)` helper. This follows the existing `op.Scopes(Configuration)` / `Config.SupportedScopes` pattern, so it is an additive field on the `Config` struct rather than a new method on the `Configuration` interface. It defaults to `false` and is omitted from the discovery document.

### 2. `feat(op): support resource indicators (RFC 8707) at the token endpoint`

Covers [RFC 8707, section 2.2](https://www.rfc-editor.org/rfc/rfc8707#section-2.2), which is what lets a client narrow the audience of the token it actually receives.

- `Resource` is added to `oidc.AccessTokenRequest`, `oidc.RefreshTokenRequest` and `oidc.ClientCredentialsRequest`.
- `op.ValidateTokenRequestResources` checks the syntax and, in addition, that every requested resource was granted by the original authorization request. A resource that was not granted is rejected with `invalid_target`.
- Because the library cannot know how a resource maps onto an audience, the values are handed to the `Storage` implementation through two **optional** interfaces on the request types it already returns:
  - `op.ResourceRequest` (`GetResource() []string`) reports the granted resources;
  - `op.CurrentResourceSetter` (`SetCurrentResources([]string)`) receives the narrowed set before the tokens are created.

  This mirrors the existing `SetCurrentScopes` on `op.RefreshTokenRequest`. An implementation that implements neither keeps its current behaviour exactly.
- `ValidateAuthReqResources` from the first commit is renamed to `ValidateResourceIndicators`, since it is no longer specific to the authorization request.
- While wiring this up I noticed that the authorization path of the new `Server` API (`webServer.authorize`) mirrors `ValidateAuthRequestClient` but never called the resource validation, so the first commit only validated on the legacy path. That is fixed here.

### 3. `feat(op): support resource indicators (RFC 8707) in the device flow`

- `Resource` is added to `oidc.DeviceAuthorizationRequest` and `oidc.DeviceAccessTokenRequest`, and both are validated.
- `op.DeviceAuthorizationState` carries the values and implements `ResourceRequest` and `CurrentResourceSetter`, so a device access token request can narrow the granted resources down.
- Storing the resources needs an extra argument, which cannot be added to `DeviceAuthorizationStorage.StoreDeviceAuthorization` without breaking every implementation. The optional `op.CanStoreDeviceAuthorizationWithResources` interface is used instead when a storage implements it, following the pattern of the other optional `Can...` storage interfaces.

### 4. `feat(rp): add WithResourceURLParam option` and `feat(example): bind the token audience to the requested resources`

- `rp.WithResourceURLParam` is a `URLParamOpt`, so it applies to both the authorization request and the token request. It sets a single value: `oauth2.SetAuthURLParam` does `url.Values.Set`, so a repeated URL parameter cannot be expressed through an `oauth2.AuthCodeOption`. That limitation is stated on the option rather than silently dropping extra values.
- The example storage now adds the requested resources to the audience of the issued tokens, and carries them on the stored access and refresh tokens so that a refresh request can narrow them again. Without this the example did not show what the feature is actually for.

Tests are added in the existing table-driven style: `TestValidateResourceIndicators`, `TestValidateTokenRequestResources`, `TestParseTokenRequestResource`, `TestDeviceAuthorizationResources`, `TestCopyRequestObjectToAuthRequest`, `Test_ResourceIndicatorsSupported`, `TestWithResourceURLParam`, `TestAuthRequestResources`, `TestRefreshTokenRequestResources`, plus new cases in `TestParseAuthorizeRequest`, `TestValidateAuthRequest`, `TestParseDeviceCodeRequest` and `TestDiscover`.

# Additional Changes

- Behaviour note: a request that carries a syntactically invalid `resource` value is now rejected with `invalid_target` where it was previously ignored. This is what the RFC requires, but it is a change in behaviour for a provider whose clients currently send malformed values. If you would rather have this gated behind `Config.ResourceIndicatorsSupported`, I am happy to change it.
- `resource_indicators_supported` is not an IANA-registered authorization server metadata parameter — RFC 8707 does not define one. It is the name used by convention to advertise the capability, and this is noted in the field's doc comment. Happy to drop it if you would prefer not to emit an unregistered parameter.
- Two pre-existing tests, `TestRoutes/…` and `TestServerRoutes/…`, compare `expires_in` against `299` and fail on a fast machine when the value is still `300`. They fail the same way on an unmodified `main`, so they are untouched here.

# Deliberately Out of Scope

- A repeated-value variant of `rp.WithResourceURLParam`, which needs a change to `withURLParam` that goes around `oauth2.AuthCodeOption`.
- Token exchange already has a `Resource` field and is left alone.

# Additional Context

- Closes #794
- RFC 8707: https://www.rfc-editor.org/rfc/rfc8707
- MCP authorization spec, resource parameter: https://modelcontextprotocol.io/specification/draft/basic/authorization#resource-parameter-implementation


wim07101993 (changes requested):
Hi, thank you for the contribution and taking notice of https://github.com/zitadel/oidc/discussions/785. 

The changes look good so far. However, I think we need to have more than only accepting the resource indicators in the auth request. Right now the feature is not yet finished, so I won't be able to merge it. Maybe you can make this a stacked pr with the other changes on top? 

Once the feature is complete, I am happy to review again.

amartya-dev:
Thanks for the review — that is fair, the authorization endpoint on its own does not give anyone a usable feature. I have pushed the rest of it on top, as four commits that each build and test on their own:

1. `f6420ea` — **token endpoint** ([RFC 8707 §2.2](https://www.rfc-editor.org/rfc/rfc8707#section-2.2)): `resource` on `AccessTokenRequest`, `RefreshTokenRequest` and `ClientCredentialsRequest`, validated both for syntax and for having been granted by the original authorization request, otherwise `invalid_target`.
2. `e854fa4` — **device flow**: `resource` on the device authorization and device access token requests, carried on `op.DeviceAuthorizationState`.
3. `c360872` — **`rp.WithResourceURLParam`**.
4. `5905998` — **example storage**, so the requested resources actually reach the token audience and the example shows what the feature is for.

Two design points worth your attention, since they are the parts I would most like a second opinion on:

**Nothing breaks existing implementations.** The library cannot know how a resource maps onto an audience, so the requested values reach `Storage` through two optional interfaces on the request types it already returns: `op.ResourceRequest` (`GetResource`) reports what was granted, and `op.CurrentResourceSetter` (`SetCurrentResources`) receives the narrowed set before the tokens are created. That mirrors the existing `SetCurrentScopes` on `op.RefreshTokenRequest`. An implementation that implements neither behaves exactly as it does today.

**The device flow needed one exception.** Storing the resources requires an extra argument on `DeviceAuthorizationStorage.StoreDeviceAuthorization`, which cannot be added without breaking every implementation, so there is an optional `op.CanStoreDeviceAuthorizationWithResources` that is used when a storage implements it — following the existing `Can...` storage interfaces. If you would rather see this shape differently, it is an easy change.

One thing I found along the way: the authorization path of the new `Server` API (`webServer.authorize`) mirrors `ValidateAuthRequestClient` but never called the resource validation, so my first commit only validated on the legacy path. That is fixed in `f6420ea`.

On "stacked PR" — I was not sure whether you meant these commits on top of this one, or separate PRs. I have gone with commits on top, since separate PRs from a fork cannot actually be based on each other (GitHub wants the base branch in this repo), so they would each show a cumulative diff against `main` anyway. If you would still rather review them as four separate PRs, say the word and I will split them — the commits are already clean cut for it.

No rush on any of this, and thanks for taking the time given #785.

pkg/oidc/authorization.go
```diff
@@ -84,6 +84,19 @@ type AuthRequest struct {
...
+	// repeated to request a token that is valid at multiple resources.
+	//
+	// Each value must be an absolute URI without a fragment component; the op package
+	// validates the syntax and rejects invalid values with `invalid_target`. Whether a
+	// resource is acceptable, and how it translates into the audience of the issued
+	// tokens, is up to the Storage implementation, which receives these values as part
+	// of the auth request.
+	//
+	// [RFC 8707]: https://www.rfc-editor.org/rfc/rfc8707
+	Resource []string `json:"resource" schema:"resource"`
```
wim07101993, line 98: The RFC specifies that:

> For an authorization request sent as a JSON Web Token (JWT), such as when using the JWT Secured Authorization Request [[JWT-SAR](https://www.rfc-editor.org/info/rfc8707/#I-D.ietf-oauth-jwsreq)], a single resource parameter value is represented as a JSON string while multiple values are represented as an array of strings.

If I am not mistaking, with the current serialization, both a single and multiple resources would be rendered as an array. Can you fix?

E.g. like the `oidc.Audience` type

example/server/storage/storage.go
```diff
@@ -605,6 +605,7 @@ func (s *Storage) createRefreshToken(accessToken *Token, amr []string, authTime
 		Expiration:    time.Now().Add(5 * time.Hour),
 		Scopes:        accessToken.Scopes,
 		AccessToken:   accessToken.ID,
+		Resource:      accessToken.Resource,
```
wim07101993, line 608: I know this is only an example, but still. The spec says:

> In the authorization_code case where the requested resources are a subset of the set of resources originally granted, the authorization server will issue an access token based on that subset of requested resources, whereas any refresh token that is returned is bound to the full original grant.

The implemented behavior is different. The refresh-token will only have the resource linked to the access-token, not the authorization request.

example/server/storage/oidc.go
```diff
@@ -211,9 +232,22 @@ func (r *RefreshTokenRequest) GetAMR() []string {
 }
 
 func (r *RefreshTokenRequest) GetAudience() []string {
+	if len(r.Resource) > 0 {
```
wim07101993, line 235: This gives an empty grant access to any resource. That sounds like a security issue. We should differentiate between "no resources granted" and "feature was not used"

example/server/storage/oidc.go
```diff
@@ -73,7 +78,22 @@ func (a *AuthRequest) GetAMR() []string {
 }
 
 func (a *AuthRequest) GetAudience() []string {
-	return []string{a.ApplicationID} // this example will always just use the client_id as audience
+	return audienceFromResources(a.ApplicationID, a.Resource)
```
wim07101993, line 81: This function is called by both the access_token and the id_token. The access token should contain resources but ID token shoudn't. 

The access-token audience change is the purpose of RFC 8707. The ID-token change is not: an ID token is issued to the client, while the API/resource server is not an intended consumer of it. OIDC permits additional ID-token audiences in some circumstances, so this is not necessarily an invalid token-`azp` is also set, but it gives those additional audiences unintended semantics and may surprise relying parties.

wim07101993 (changes requested):
Thank you for the contribution. It took some time to go over the pr but here is a first review.

I know some comments are in the example, but I think it should be correct as well.

Please also adjust the title so that it is clear that not only the authorization endpoint has resource indicators now.

FYI: Stacked PRs are a feature of github in which you can base a branch of another branch and create a series of prs which are "linked" to each other. (https://docs.github.com/en/pull-requests/how-tos/stacked-pull-requests). However, one big pr works just as well for me.

PR #959, opened 2026-08-25
fix(op): guard against nil claims for opaque access tokens in token exchange

### LINKED ISSUE #704: [Bug]: null pointer exception in GetTokenIDAndSubjectFromToken
### Preflight Checklist

- [x] I could not find a solution in the documentation, the existing issues or discussions

### Version

3.34.1 and master branch

In `pkg/op/exchange_token.go`:
From [L309-313](https://github.com/zitadel/oidc/blob/8c9a5360587d988691f7b27ed1c2fb9d5e49fc00/pkg/op/token_exchange.go#L309-L313):
```
tokenIDOrToken, subject, accessTokenClaims, ok = getTokenIDAndClaims(ctx, exchanger, token)
		if !ok {
			break
		}
		claims = accessTokenClaims.Claims
```

And the `getTokenIDAndClaims` function has a return case at  [L424](https://github.com/zitadel/oidc/blob/8c9a5360587d988691f7b27ed1c2fb9d5e49fc00/pkg/op/token_exchange.go#L424):
```
return splitToken[0], splitToken[1], nil, true
```

When this codepath is hit, ok is true, then the `accessTokenClaims` is a nil point, which is accessed.

Second bug: this usecase is not covered by unit tests.

### To reproduce

Try to do a token exchange which is hitting this codepath (user initiates a client credential grant flow, then tries to exchange the token to have the audience match the resource provider it is trying to access).

### Screenshots

_No response_

### Expected behavior

The token exchange working properly.

### Additional Context

Discord gladly has thrown some errors which thankfully made me unable to join.

ChrisJr404:
Fixes #704.

Doing a token exchange with an opaque access token panics. When the token is an encrypted tokenID:subject string rather than a JWT, getTokenIDAndClaims returns nil claims with ok true, and GetTokenIDAndSubjectFromToken then dereferences that nil pointer on accessTokenClaims.Claims.

I added a nil guard so opaque tokens come back with empty claims (there are no JWT claims to read in that case), and a test for the opaque path since, as the issue mentions, it wasn't covered before.


wim07101993 (approved):

wim07101993:
Thank you for the contribution. Will merge.

PR #963, opened 2026-08-27
fix(client): avoid mutating caller http.Client in CallRevokeEndpoint

### LINKED ISSUE #911: [Bug]: `client.CallRevokeEndpoint` mutates httpClient
### Preflight Checklist

- [x] I could not find a solution in the documentation, the existing issues or discussions

### Version

v3.47.5

Calling `client.CallRevokeEndpoint` with an custom httpClient mutates the clients which lead to an unexpected behavior for the method. 

https://github.com/zitadel/oidc/blob/c9fca57a5d85c0a9ab96ce838be296137899aef5/pkg/client/client.go#L182-L184

It may results to unknown side effects, if the httpClient is shared with other areas.

### To reproduce

1. Call `client.CallRevokeEndpoint `

### Screenshots

_No response_

### Expected behavior

_No response_

### Additional Context

_No response_

AhmedShehab1:
# Which Problems Are Solved
- Calling `CallRevokeEndpoint` or `CallEndSessionEndpoint` directly mutated the caller's `*http.Client.CheckRedirect` field, permanently stopping the client from following redirects for subsequent HTTP calls and introducing potential data races when shared across goroutines.

# How the Problems Are Solved
- Made a shallow copy (`cl := *client`) prior to setting `CheckRedirect = http.ErrUseLastResponse`, ensuring the caller's `*http.Client` remains untouched while still sharing the underlying `http.Transport`.
- Added unit tests in `pkg/client/client_test.go` covering both endpoints with `nil` and custom `CheckRedirect` handlers.

# Additional Changes
- None

# Additional Context
- Closes #911

### DIFF
--- a/pkg/client/client.go
+++ b/pkg/client/client.go
@@ -116,10 +116,11 @@ func CallEndSessionEndpoint(ctx context.Context, request any, authFn any, caller
 		return nil, err
 	}
 	client := caller.HttpClient()
-	client.CheckRedirect = func(_ *http.Request, _ []*http.Request) error {
+	cl := *client
+	cl.CheckRedirect = func(_ *http.Request, _ []*http.Request) error {
 		return http.ErrUseLastResponse
 	}
-	resp, err := client.Do(req)
+	resp, err := cl.Do(req)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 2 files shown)


wim07101993 (approved):
Thanks for the contribution. Will merge

PR #775, opened 2025-07-20
fix: add redirect_uri decoded

Sianao:
### Definition of Ready

This PR introduces a redirect_uri decoding step (url.QueryUnescape) in the authorization request validation logic. 
Libraries such as [golang.org/x/oauth2](https://cs.opensource.google/go/x/oauth2/+/refs/tags/v0.30.0:oauth2.go;l=184) automatically encode the redirect_uri using url.Values.Encode(). This means the incoming URI is percent-encoded (e.g., https%3A%2F%2Fclient.example.com%2Fcallback), and the server must decode it before performing string comparisons.

- [x] I am happy with the code
- [x] Short description of the feature/issue is added in the pr description
- [x] PR is linked to the corresponding user story
- [x] Acceptance criteria are met
- [x] All open todos and follow ups are defined in a new ticket and justified
- [x] Deviations from the acceptance criteria and design are agreed with the PO and documented.
- [x] No debug or dead code
- [x] My code has no repetitions
- [x] Critical parts are tested automatically
- [x] Where possible E2E tests are implemented
- [x] Documentation/examples are up-to-date
- [x] All non-functional requirements are met
- [x] Functionality of the acceptance criteria is checked manually on the dev system.

PR #969, opened 2026-09-02
fix(op): stop double-decoding redirect_uri in authorization requests

### LINKED ISSUE #968: [Bug]: ValidateAuthReqRedirectURI double-decodes redirect_uri, rejecting registered URIs containing `+` or `%XX` (regression in v3.43.1)
## Summary

`op.ValidateAuthReqRedirectURI` (`pkg/op/auth_request.go:338`) runs `url.QueryUnescape` on `redirect_uri` before matching it against `client.RedirectURIs()`. The value it receives has already been percent-decoded by `r.ParseForm()` in `decodeRequest` (`pkg/op/server_http.go:511-524`). The second decode alters any registered redirect URI that contains a `+` or a `%XX` escape, so clients with such a URI are refused since v3.43.1. Present on `main` today.

## Where

- `pkg/op/server_http.go:203-215` — `authorizeHandler` calls `decodeRequest[oidc.AuthRequest](s.decoder, r, false)`.
- `pkg/op/server_http.go:511-524` — `decodeRequest` calls `r.ParseForm()` and decodes the struct from `r.Form`. This is the single correct decode; from here `authReq.RedirectURI` holds the URI exactly as the client registered it.
- `pkg/op/auth_request.go:336-364` — `ValidateAuthReqRedirectURI`. Line 338, `uri, err := url.QueryUnescape(uri)`, decodes a second time into a local shadow. `authReq.RedirectURI` is not modified, so the redirect eventually issued uses the once-decoded value while the match uses the twice-decoded one.
- `pkg/op/auth_request.go:314-334` — `checkURIAgainstRedirects`, `slices.Contains(client.RedirectURIs(), uri)`: exact match consuming the twice-decoded value.
- Both entry points are affected: `webServer.authorize` (`pkg/op/server_http.go:217-240`, `op.Server` path, line 234) and `ValidateAuthRequestClient` (`pkg/op/auth_request.go:260-278`, Provider path, line 264).

## Origin

#775 "fix: add redirect_uri decoded", merged 2025-08-04, released in v3.43.1. Its premise:

> Libraries such as golang.org/x/oauth2 automatically encode the redirect_uri using url.Values.Encode(). This means the incoming URI is percent-encoded (e.g., https%3A%2F%2Fclient.example.com%2Fcallback), and the server must decode it before performing string comparisons.

`url.Values.Encode()` is the wire encoding. `net/http` undoes it in `r.ParseForm()`, which `decodeRequest` calls before the struct decoder runs. By the time `ValidateAuthReqRedirectURI` sees the value it is `https://client.example.com/callback`, never `https%3A%2F%2F...`.

Checked against `golang.org/x/oauth2` itself: `Config.AuthCodeURL` does `v.Set("redirect_uri", c.RedirectURL)` and later `buf.WriteString(v.Encode())` (`oauth2.go:167` and `:183` at v0.36.0; `:168` and `:184` at v0.30.0, the version #775 links — its "line 184" is the `v.Encode()` write). `Set` stores `RedirectURL` verbatim; `oauth2.go` has never applied `QueryEscape` to it (checked at v0.1.0, v0.7.0, v0.13.0, v0.21.0, v0.30.0, v0.36.0). x/oauth2 percent-encodes `redirect_uri` exactly once, as every OAuth client must.

## Why the test added in #775 passes

`pkg/op/auth_request_test.go:733-740`, "code flow encoded redirect_uri should pass after unescaping", calls `op.ValidateAuthReqRedirectURI` directly with the literal string `"https%3A%2F%2Fregistered.com%2Fcallback"`. No HTTP request ever produces that string at that call site, because `ParseForm` has already decoded it. The test asserts the workaround against its own assumption instead of driving a request through the handler, so it cannot observe the double decode.

## Consequence

Not an authentication bypass and not an open redirect: the comparison is still an exact `slices.Contains` against the registered list, and the extra decode can only move the compared value away from what the client sent. The effect is false rejection of legitimately registered redirect URIs:

- Registered `http://localhost:9999/cb?a=b+c`. The client encodes once, so `ParseForm` yields `http://localhost:9999/cb?a=b+c`. `QueryUnescape` then turns `+` into a space: `http://localhost:9999/cb?a=b c`, which is not in `RedirectURIs()`.
- Any registered URI containing `%2F`, `%25`, `%3D`, … is decoded one level further and stops matching the same way.

Such clients worked before v3.43.1 and are refused after it.

## Reproduction

Measured against the bundled example server, which registers its `web` client with the comma-separated `REDIRECT_URI` env var (`example/server/main.go:34`, `example/server/config/config.go:36`) and mounts the authorization endpoint at `/auth` (`example/server/exampleop/op.go:159`).

Register one ordinary redirect URI and one containing a `+`, on the same client:

```sh
REDIRECT_URI='http://localhost:9999/plain,http://localhost:9999/cb?a=b+c' \
  go run github.com/zitadel/oidc/v3/example/server@v3.49.4
```

`--data-urlencode` encodes once, producing the same bytes `x/oauth2` would send:

```sh
# control — ordinary registered URI
curl -s -o /dev/null -w '%{http_code} %{redirect_url}\n' -G http://localhost:9998/auth \
  --data-urlencode 'client_id=web' --data-urlencode 'response_type=code' \
  --data-urlencode 'scope=openid' \
  --data-urlencode 'redirect_uri=http://localhost:9999/plain'

# subject — registered URI containing '+'
curl -s -w '%{http_code}\n' -G http://localhost:9998/auth \
  --data-urlencode 'client_id=web' --data-urlencode 'response_type=code' \
  --data-urlencode 'scope=openid' \
  --data-urlencode 'redirect_uri=http://localhost:9999/cb?a=b+c'
```

Observed on v3.49.4:

```
302 http://localhost:9998/login/username?authRequestID=cd38c024-acb4-4f75-8290-daa586227baf
400
```

The 400 body is `The requested redirect_uri is missing in the client configuration. …`, i.e. `checkURIAgainstRedirects` did not find the registered URI. Substituting a registered `http://localhost:9999/cb?p=a%2Fb` and requesting it gives the same 400.

Across versions, with only the `+` URI registered:

- v3.43.0 — `302 Found`, `Location: /login/username?authRequestID=…`
- v3.49.4 — `400 Bad Request`, the message above

Removing line 338 of `pkg/op/auth_request.go` restores the redirect.

## Proposed fix

Revert #775. If a regression test is wanted, drive it through `RegisterServer` / the HTTP handler with `httptest` so `ParseForm` participates; a direct call to `ValidateAuthReqRedirectURI` can only reproduce the assumption the existing test was written under.

ElectronicRU:
# Which Problems Are Solved

- Since v3.43.1, a client whose registered redirect URI contains a `+` or a `%XX` escape (for example `https://app.example/cb?p=a%2Fb`) is refused with `invalid_request` ("The requested redirect_uri is missing in the client configuration") even though it sends exactly the registered value.
- `ValidateAuthReqRedirectURI` calls `url.QueryUnescape` on a `redirect_uri` that `r.ParseForm` in `decodeRequest` has already decoded once. The second decode alters the value, and the exact `slices.Contains` match against `client.RedirectURIs()` fails.
- Both entry points are affected: `webServer.authorize` (`op.Server`) and `ValidateAuthRequestClient` (Provider).

# How the Problems Are Solved

- Removes the `url.QueryUnescape` call introduced by #775. Its premise, that `golang.org/x/oauth2` sends `redirect_uri` percent-encoded and the server must undo it, does not hold: `url.Values.Encode` is the wire encoding, `net/http` reverses it in `ParseForm`, and `Config.AuthCodeURL` sets the raw `RedirectURL` and encodes it once (no `QueryEscape` in any release from v0.1.0 to v0.36.0).
- Adds `Test_webServer_authorizeHandler_encodedRedirectURI`, which drives `authorizeHandler` over HTTP via `httptest` with a query built by `url.Values.Encode`, for a registered URI containing `%2F` and one containing `+`, and expects a 302 to the client login URL. The test in #775 called `ValidateAuthReqRedirectURI` directly, so `ParseForm` never ran and a double decode was unobservable; only an HTTP-level test can pin this down.

# Additional Changes

- `TestValidateAuthReqRedirectURI`: the #775 case is inverted. A percent-encoded literal must not match the registered URI, so re-adding a decode fails a test instead of passing one.
- `testClient.RedirectURIs()` gains the two encoded entries; `redirectingAuthorizer` implements `Authorize` so successful validation shows up as a redirect rather than an unimplemented-method error.
- The table loop in `TestValidateAuthReqRedirectURI` printed `err.Error()`, which panics on a nil error whenever a `wantErr: true` case wrongly succeeds. It now prints the error value.

# Additional Context

- Closes #968
- Reverts #775
- #968 holds the full analysis, per-version measurements and a curl reproduction against the example server.
- This is a false-rejection regression, not a security issue: the extra decode could only move the compared value away from what the client sent, never toward an unregistered URI.

### DIFF
--- a/pkg/op/auth_request.go
+++ b/pkg/op/auth_request.go
@@ -335,8 +335,7 @@ func checkURIAgainstRedirects(client Client, uri string) error {
 
 // ValidateAuthReqRedirectURI validates the passed redirect_uri and response_type to the registered uris and client type
 func ValidateAuthReqRedirectURI(client Client, uri string, responseType oidc.ResponseType) error {
-	uri, err := url.QueryUnescape(uri)
-	if uri == "" || err != nil {
+	if uri == "" {
 		return oidc.ErrInvalidRequestRedirectURI().WithDescription("The redirect_uri is missing in the request. " +
 			"Please ensure it is added to the request. If you have any questions, you may contact the administrator of the application.")
 	}
--- a/pkg/op/auth_request_test.go
+++ b/pkg/op/auth_request_test.go
@@ -731,19 +731,23 @@ func TestValidateAuthReqRedirectURI(t *testing.T) {
 			true,
 		},
 		{
-			"code flow encoded redirect_uri should pass after unescaping",
+			// A redirect_uri reaches this validator already decoded by
+			// r.ParseForm, so a percent-encoded literal denotes a
+			// different URI than the one registered and must not match.
+			// Guards against re-introducing a second decode here; see #968.
+			"code flow percent-encoded redirect_uri does not match",
 			args{
 				"https%3A%2F%2Fregistered.com%2Fcallback",
 				mock.NewClientWithConfig(t, []string{"https://registered.com/callback"}, op.ApplicationTypeWeb, nil, false),
 				oidc.ResponseTypeCode,
 			},
-			false,
+			true,
 		},
 	}
 	for _, tt := range tests {
 		t.Run(tt.name, func(t *testing.T) {
 			if err := op.ValidateAuthReqRedirectURI(tt.args.client, tt.args.uri, tt.args.responseType); (err != nil) != tt.wantErr {
-				t.Errorf("ValidateRedirectURI() error = %v, wantErr %v", err.Error(), tt.wantErr)
+				t.Errorf("ValidateRedirectURI() error = %v, wantErr %v", err, tt.wantErr)
 			}
 		})
 	}
--- a/pkg/op/server_http_test.go
+++ b/pkg/op/server_http_test.go
@@ -96,6 +96,9 @@ func (c *testClient) RedirectURIs() []string {
 		"http://registered.com/callback",
 		"http://localhost:9999/callback",
 		"custom://callback",
+		// Survive one percent-decode unchanged, but not two.
+		"https://registered.com/callback?p=a%2Fb",
+		"https://registered.com/callback?a=b+c",
 	}
 }
 
@@ -191,6 +194,17 @@ func (s *requestVerifier) VerifyClient(ctx context.Context, r *Request[ClientCre
 		return nil, oidc.ErrServerError()
 	}
 	return s.client, nil
+}
+
+// redirectingAuthorizer verifies like requestVerifier and then completes the
+// authorization request the way a real Server implementation would, so that a
+// test can tell successful validation apart from any error response.
+type redirectingAuthorizer struct {
+	requestVerifier
+}
+
+func (s *redirectingAuthorizer) Authorize(ctx context.Context, r *ClientRequest[oidc.AuthRequest]) (*Redirect, error) {
+	return NewRedirect(r.Client.LoginURL("authReqID")), nil
 }
 
 var testDecoder = func() *schema.Decoder {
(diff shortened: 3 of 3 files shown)


wim07101993 (approved):
Thank you for the contribution, will merge.

PR #971, opened 2026-09-05
feat(rp): limit HTTP response body size

jkroepke:
# Which Problems Are Solved

This PR limits the maximum size of HTTP response bodies before reading them into memory.

Previously, `io.ReadAll(resp.Body)` could read an unbounded response body. The response body is now wrapped with `io.LimitReader` using `MaxResponseBodySize`.

# How the Problems Are Solved

This reduces the risk of excessive memory usage caused by unexpectedly large HTTP responses which can lead into potential DoS attack.

# Additional Changes

N/A

# Additional Context

N/A


pkg/http/const.go
```diff
@@ -0,0 +1,6 @@
+package http
+
+const (
+	// MaxResponseBodySize HTTP responses are limited to 1 MiB.
+	MaxResponseBodySize = 1 << 20
```
wim07101993: Maybe make it a var, so that it is configurable. Most responses will never get to 1MiB so I don't seen anyone running into problems. But it is nice to have the option.

pkg/http/http.go
```diff
@@ -78,6 +78,10 @@ func HttpRequest(client *http.Client, req *http.Request, response any) error {
 		return &oidcErr
 	}
 
+	if len(body) > MaxResponseBodySize {
+		return fmt.Errorf("http response too large")
```
wim07101993: Can you make the error a package level sentinel?

wim07101993 (changes requested):
Hi, thank you for the contribution. Just a few small comments.

pkg/http/const.go
jkroepke: As global var?

wim07101993: I think that is the easiest. Otherwise we would have to change the contract of the function.

jkroepke: Done. `MaxResponseBodySize` is now an exported `int64` variable with a 1 MiB default, so callers can configure it without changing the `HttpRequest` API.

pkg/http/http.go
jkroepke: Done. Added package-level sentinel `ErrResponseBodyTooLarge` and return it when the configured response body limit is exceeded.

jkroepke:
Done, the commits was generated by Codex, thats why they remain unverified.

pkg/http/http.go
```diff
@@ -78,6 +78,10 @@ func HttpRequest(client *http.Client, req *http.Request, response any) error {
 		return &oidcErr
 	}
 
+	if int64(len(body)) > MaxResponseBodySize {
```
wim07101993, line 81: Shouldn't this check happen before the status check? Right now a call can fail with a json unmarshalling because of an incomplete json becaues the limitreader truncated half of the response

wim07101993 (commented):
Sorry, after going over it, I found one more thing.

pkg/http/http.go
jkroepke: I would not agree here.

If there is a 500 error with a response that is too large, it would be more helpful to see the truncated response than a generic "body too large" error.

The code already has a path for JSON unmarshalling failures and simply returns the truncated body as the error message.

wim07101993: fair

wim07101993 (approved):

PR #979, opened 2026-09-17
fix(op): stop double-escaping fragment response parameters

Mohil-Ahuja:
# Which Problems Are Solved

- When the authorization response is sent with `response_mode=fragment`, or as the implicit flow's default fragment, parameters are percent-encoded twice. A `state` of `PADTEST==` comes back as `PADTEST%253D%253D`, which decodes to `PADTEST%3D%3D` instead of the value the client sent, so clients that check `state` drop the response.

# How the Problems Are Solved

- `setFragment` assigned the already encoded output of `params.Encode()` to `url.URL.Fragment`, which is meant to hold the decoded fragment, so `String()` escaped it again. It now sets `RawFragment` to the encoded parameters and `Fragment` to their decoded form, so `String()` emits the encoded version once. This matches how `mergeQueryParams` writes `RawQuery` for query mode, which was not affected.

# Additional Changes

- Two cases in `TestAuthResponseURL` with values that need escaping (`=` padding, space, `/`, `&`), one for explicit fragment mode and one for the id token default. Both fail on `main` with the doubled `%25` and pass with the change. The existing fragment cases are unchanged.

# Additional Context

- Reported against ZITADEL in zitadel/zitadel#12645, which reproduces it with a base64 `state` on 4.17.1 and 4.16.0.


wim07101993 (approved):

wim07101993:
Thank you for the contribution. Will merge

PR #981, opened 2026-09-17
fix(op): escape the request id in the auth callback URL

Mohil-Ahuja:
# Which Problems Are Solved

- `AuthCallbackURL` builds the post-login redirect as `...authorize/callback?id=` + the request id, without escaping it. The callback handler reads the id back with `r.Form.Get("id")`, which decodes it. For a storage whose auth request ids are not plain alphanumerics (base64, URL-safe tokens, compound keys) the id comes back different: `+` turns into a space, `&` or `#` cut it short, and a `%` makes the URL fail to parse. The callback then cannot find the auth request. ZITADEL's own numeric ids are not affected.

# How the Problems Are Solved

- The id is passed through `url.QueryEscape` in both builders, `op.AuthCallbackURL` and `LegacyServer.AuthCallbackURL`.

# Additional Changes

- `TestAuthCallbackURL_EscapesRequestID` builds the callback URL through both builders with the id `a+b/c=d&e#f %`, parses it, and checks `Query().Get("id")` returns the id unchanged. Both cases fail on `main`.

# Additional Context

- Same family as #979 (fragment response parameters escaped twice): values placed into URLs by hand rather than through `url.URL` / `url.Values`.


wim07101993 (approved):

wim07101993:
Thank you for the contribution. Will merge

